PR #1, opened 2019-10-14
services: Add POSIX fs support

Xuanwo:
This will add full POSIX support for storage.

### DIFF
--- a/types/struct.go
+++ b/types/struct.go
@@ -6,9 +6,9 @@ type ObjectType string
 // All available type for object.
 const (
 	ObjectTypeFile    ObjectType = "file"
-	ObjectTypeStream             = "stream"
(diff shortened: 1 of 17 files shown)


types/struct.go
```diff
@@ -1,18 +1,22 @@
...
 
+// ObjectType is the type for object, under layer type is string.
+type ObjectType string
+
 // All available type for object.
 const (
-	ObjectTypeFile   = "file"
-	ObjectTypeStream = "stream"
-	ObjectTypeDir    = "dir"
+	ObjectTypeFile    ObjectType = "file"
```
golangcibot: SA9004: only the first constant in this group has an explicit type (from `staticcheck`)

PR #11, opened 2019-11-02
service/posixfs/storager: Modify output key with relative path in front

Prnyself:


Xuanwo:
Fixed in https://github.com/Xuanwo/storage/pull/11

PR #56, opened 2020-01-03
services,types: Implement proposal add-id-in-object

Xuanwo:


Xuanwo:
Fixed https://github.com/Xuanwo/storage/issues/54

PR #136, opened 2020-02-27
services/qingstor: Add tests for errors not handled as intended

Prnyself:


services/qingstor/storager_test.go
```diff
@@ -891,3 +891,85 @@ func TestStorage_ListSegments(t *testing.T) {
...
+	for _, tt := range cases {
+		err := s.formatError(tt.op, tt.err, tt.path)
+		if tt.err == nil {
+			assert.Nil(t, err, tt.name)
+			continue
+		}
+
+		e, ok := err.(*services.StorageError)
+		assert.True(t, ok, tt.name)
+		assert.Equal(t, tt.op, e.Op, tt.name)
```
Xuanwo: We will not support direct type assert use case any more, so we don't need to test this.

Instead, we need to check if `errors.Is` and `errors.As` works as expected.

PR #223, opened 2020-03-24
services/qingstor: Dir support is not needed in stat

Xuanwo:


Xuanwo:
We don't need to take care about directory support in stat, we just return what we get from underlying storage.

PR #484, opened 2020-12-10
Proposal: Add default pair for operations

Prnyself:


docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,52 @@
...
+and we should support default pairs, for different operations. 
+Related issue: [proposal: Add default set for operations #170]
+
+## Proposal
+
+So I propose that [storage] should support default pairs for different operations:
+
+### Add default pairs
+
+We can set a struct for every supported operation in [storage].
```
Xuanwo, line 23: What's the struct looks like? Where we maintain the struct? How we miantain the strcut?

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,52 @@
...
+So that in different service, it can implement its custom pairs when init.
+
+### Parse pairs in specific operation
+
+When parsing pairs in specific operation, we should combine default pairs and `pairs from args`,
```
Xuanwo, line 28: How?

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,52 @@
...
+we call specific operation.
+
+## Compatibility
+
+None
+
+## Implementation
+
+```go
+type DefaultPairs struct {
```
Xuanwo, line 43: How to handle the conflict name between services and storager?

Like: both service and storage have `delete`

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,52 @@
...
+Related issue: [proposal: Add default set for operations #170]
+
+## Proposal
+
+So I propose that [storage] should support default pairs for different operations:
+
+### Add default pairs
+
+We can set a struct for every supported operation in [storage].
+So that in different service, it can implement its custom pairs when init.
```
Xuanwo, line 24: What's the behavior if the default pairs are not valid?

docs/design/22-add-default-set-for-operations.md
Xuanwo: DefaultPairs or FallbackPairs?

docs/design/22-add-default-set-for-operations.md
Prnyself: You mean value not valid?

Xuanwo: Yes.

Will we check the pairs during service init?

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,104 @@
...
+This struct is maintained by different service, which defines its own operations, and should be
+generated from specs as same as definitions of operations.
+
+### Handle conflict
+
+If both `Service` and `Storage` have operations with duplication of name, such as delete,
+we should handle conflict.
+Since the operations with the same name handle different targets (`Servicer` and `Storager`),
+I think we should handle the separately.
+```go
```
Xuanwo, line 34: Add empty line here.

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,104 @@
...
+type DefaultServicePairs struct {
+    Delete []Pair
+}
+```
+
+### Define struct separately
+
+Since one service may just implement a part of operations in [storage],
+there are two ways to define default pairs struct:
+- Unify struct in [storage], so that we can maintain it only once.
```
Xuanwo, line 51: Add empty line here.

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,104 @@
...
+For example, the [fs-service] does not implement `WriteIndexSegment` operation, we do not need a field like `WriteIndexSegmentPair`.
+All the pairs are generated from specs, so we choose the second way.
+
+### Parse pairs in specific operation
+
+When parsing pairs in specific operation, we should combine default pairs and `pairs from args`,
+and make sure that `pairs from args` can overwrite default pair, and this should be **generated**.
+For example, we parse the pairs sequentially from the slice of pairs,
+so we should create a new slice of pairs and append `pairs from args` to the slice.
+```go
```
Xuanwo, line 64: Add empty line here.

docs/design/22-add-default-set-for-operations.md
```diff
@@ -0,0 +1,104 @@
...
+// generated code in Write operation
+func (s *Storage) WriteWithContext(ctx context.Context, path string, r io.Reader, pairs ...Pair) (n int64, err error) {
+    defer func() {
+        err = s.formatError("write", err, path)
+    }()
+    ps := getWritePairsFromStorage(s)
```
Xuanwo, line 70: Maybe we can generate code directly here?

Xuanwo: Something like `pairs = append(pairs, s.defaultPairs.Write...)`

With this update, we need to make sure `parsePairStorageWrite` only take the first occur pairs' value.

Prnyself: OK, we can check if the value was set before.
```go
for _, v := range opts {
	switch v.Key {
	// Required pairs
	case "name":
		if result.HasName {
			continue
		}
		result.HasName = true
		result.Name = v.Value.(string)
	}
}
```

Xuanwo (approved):

PR #518, opened 2021-03-18
cmd/definitions: Add tests service

Xuanwo:


cmd/definitions/tests/.gitignore
```diff
@@ -0,0 +1 @@
+generated.go
```
Prnyself, line 1: Why set `generated.go` ignored?

Xuanwo: `generated.go` is only used for testing and it's content is meanless, so that I think we can ignore it.

Prnyself (approved):
LGTM

PR #541, opened 2021-04-23
cmd/definitions: Implement AOS-40

Xuanwo:


Xuanwo:
All service should update their usage of service metadata before this PR get merged.

PR #555, opened 2021-05-06
Implement AOS-47 Additional Error Specification 

xxchan:
part of https://github.com/aos-dev/go-storage/issues/558


services/error.go
```diff
@@ -17,6 +20,10 @@ var (
 	ErrObjectNotExist = errors.New("object not exist")
 	// ErrPermissionDenied means this operation doesn't have enough permission.
 	ErrPermissionDenied = errors.New("permission denied")
+	// ErrObjectNotPart means the object to be operated is not a part object.
+	ErrObjectNotPart = errors.New("object is not a part object")
```
Xuanwo, line 24: We do have `not part`, `not append`, `not page`, `not read`, they are all `invalid object mode`.

Maybe we can use the same error code here?

services/error.go
```diff
@@ -17,6 +20,10 @@ var (
...
+	// ErrInvalidListMode means the provided list mode is invalid.
+	ErrInvalidListMode = errors.New("invalid list mode")
```
Xuanwo, line 26: Please use the style like `<noun><adj>`. I prefer `ListModeInvalid` instead of `InvalidListMode`.

We can discuss the error code style in AOS-47.

services/error.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 var (
+	// ErrUnexpected means this is an unexpected error which go-storage can't handle
+	ErrUnexpected = errors.New("go-storage can't handle this error")
```
Xuanwo, line 12: Not synced with AOS-47.

services/error.go
```diff
@@ -73,88 +78,76 @@ type StorageError struct {
...
+}
+
+// ObjectModeInvalidError means the provided object mode is invalid.
+type ObjectModeInvalidError struct {
+	Expected types.ObjectMode
+	Actual   types.ObjectMode
+}
+
+func (e ObjectModeInvalidError) Error() string {
+	return fmt.Sprintf("object mode invalid, expected %b, actual %b", e.Expected, e.Actual)
```
Xuanwo, line 143: `ObjectMode` && `ListMode` should implement `String() string` so that we can print human readable error here.

services/error.go
```diff
@@ -73,88 +78,76 @@ type StorageError struct {
...
 
 // Unwrap implements xerrors.Wrapper
-func (e *PairRequiredError) Unwrap() error {
-	return e.Err
+func (e PairRequiredError) Unwrap() error {
+	return ErrRestrictionDissatisfied
+}
+
+// ObjectModeInvalidError means the provided object mode is invalid.
+type ObjectModeInvalidError struct {
```
Xuanwo, line 137: It's a bit wired for an error struct that can't unwrap to an error code.

- Makes it hard for the user to discover which error go-storage could return.
- Makes it hard to check the error code: the user needs to use `errors.Is` & `errors.As` at the same time. What if user just wants to know what error happened and doesn't care about why and related context?

Maybe we should provide an error code for error struct (even when they are already the most underlying error)?

xxchan: I agree. If every error has an error code, it will be easier to understand.
Just like HTTP response codes can have 0, 1 or more error codes, our error codes can have 0, 1 or more error `struct`s?

Xuanwo: Yes, our error codes could be used as underlying errors for different error structs.

- error code represents `What` -> only a label to explain why the error happened
- error struct represents `Why` -> carrying context info to explain why the error happened

Xuanwo (approved):

PR #564, opened 2021-05-10
Implement AOS-51: Distinguish Errors by IsAosError

xxchan:
https://github.com/aos-dev/specs/pull/51

### DIFF
--- a/services/error.go
+++ b/services/error.go
@@ -1,31 +1,57 @@
 package services
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/aos-dev/go-storage/v3/types"
 )
 
+type AosError interface {
+	// IsAosError SHOULD and SHOULD ONLY be implemented by error definitions in go-storage & go-service-*.
+	// We depends on the AosError interface to distinguish our errors.
+	// There's no need for user code to implement or use this function and interface.
+	IsAosError()
+}
+
+// Create a new error code.
+//
+// Developers SHOULD use this function to define error codes (sentinel errors), instead of `NewErrorCode`
+//
+// Users SHOULD NOT call this function. Use defined error codes instead.
+func NewErrorCode(text string) error {
+	return errorCode{text}
+}
+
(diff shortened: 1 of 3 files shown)


services/error.go
```diff
@@ -1,31 +1,57 @@
...
+	// There's no need for user code to implement or use this function and interface.
+	IsAosError()
+}
+
+// Create a new error code.
+//
+// Developers SHOULD use this function to define error codes (sentinel errors), instead of `NewErrorCode`
+//
+// Users SHOULD NOT call this function. Use defined error codes instead.
+func NewErrorCode(text string) error {
```
Xuanwo, line 21: How about `type ErrorCode string`, so that we can

```
ErrUnexpected = ErrorCode("unexpected")
```

I think the `ErrorCode` type should be exported so that services can also use them?

xxchan: I think both are ok. No big difference.

The reason why I hide it is that I think this type will never be used (also for services)

Xuanwo: > Perhaps letting `ErrUnexpected`'s type be `error` can cause less cognitive load for users?

I prefer returning the same type for all error codes.

> The reason why I hide it is that I think this type will never be used (also for services)

Service may need to define error like `ErrInvalidBukcet`? I think those error should also use our ErrorCode type.

xxchan: > Service may need to define error like `ErrInvalidBukcet`? I think those error should also use our ErrorCode type.

That's why I provide the `NewErrorCode` constructor function.

Xuanwo: OK, sounds reasonable, accepted.

xxchan: 😂 Wait, so what's the final choice

Xuanwo: Current implementations LGTM.

xxchan:
Why do we need to update dependences? Its code does not change

Xuanwo (approved):

PR #573, opened 2021-05-12
Implement AOS-61 Add object mode check for operations

Prnyself:
Signed-off-by: Prnyself <281239768@qq.com>


Prnyself:
ref: aos-dev/go-storage#557

cmd/definitions/tmpl/service.tmpl
```diff
@@ -306,6 +303,11 @@ func (s *{{ $pn }})parsePair{{ $pn }}{{ $fnk }}(opts []Pair) (pair{{ $pn }}{{ $f
             err = s.formatError("{{$fn.Name}}", err {{ $path }} )
         }()
 
+        {{- template "mode_check" makeSlice $fn.ObjectMode $fn.Params.ObjectParamName }}
```
Xuanwo, line 306: How about expanding the template here? It adds an extra burden to understand the logic here (introduce a new template for just three lines).

Xuanwo: And I think we don't need to check whether we have Object here.

Those functions that have object mode must have object as param.

Check `$fn.ObjectMode` is enough to me.

Prnyself: The `ObjectParamName` is for getting the param name of `Object`, since we can ensure the name of param `Object` is always `o`, maybe I should remove this func?

cmd/definitions/tmpl/service.tmpl
```diff
@@ -303,7 +303,13 @@ func (s *{{ $pn }})parsePair{{ $pn }}{{ $fnk }}(opts []Pair) (pair{{ $pn }}{{ $f
             err = s.formatError("{{$fn.Name}}", err {{ $path }} )
         }()
 
-        {{- template "mode_check" makeSlice $fn.ObjectMode $fn.Params.ObjectParamName }}
+        {{- if ne $fn.ObjectMode ""}}
+            {{- $modeP := $fn.ObjectMode | toPascal }}
+            if !{{ $fn.Params.ObjectParamName }}.Mode.Is{{ $modeP }}() {
+                err = services.ObjectModeInvalidError{Expected: Mode{{ $modeP }}, Actual: {{$fn.Params.ObjectParamName}}.Mode}
```
Xuanwo, line 309: Why we need `{{$fn.Params.ObjectParamName}}.Mode` here, can we use `o.Mode`?

cmd/definitions/tmpl/service.tmpl
```diff
@@ -303,7 +303,13 @@ func (s *{{ $pn }})parsePair{{ $pn }}{{ $fnk }}(opts []Pair) (pair{{ $pn }}{{ $f
             err = s.formatError("{{$fn.Name}}", err {{ $path }} )
         }()
 
-        {{- template "mode_check" makeSlice $fn.ObjectMode $fn.Params.ObjectParamName }}
+        {{- if ne $fn.ObjectMode ""}}
+            {{- $modeP := $fn.ObjectMode | toPascal }}
```
Xuanwo, line 307: `$mode` is good to me.

cmd/definitions/tmpl/service.tmpl
Xuanwo: Yes, Object will always be `o *Object`, it's ensured by our specs.

cmd/definitions/tmpl/service.tmpl
Prnyself: The ObjectParamName is for getting the param name of Object, since we can ensure the var of param Object is always o, maybe I should remove this func?

Xuanwo: Yes, remove this func helps improve the readability.

cmd/definitions/tmpl/service.tmpl
```diff
@@ -306,6 +303,17 @@ func (s *{{ $pn }})parsePair{{ $pn }}{{ $fnk }}(opts []Pair) (pair{{ $pn }}{{ $f
...
 
+        {{- if ne $fn.ObjectMode ""}}
+            {{- $mode := $fn.ObjectMode | toPascal }}
+            if !o.Mode.Is{{ $mode }}() {
+                err = services.ObjectModeInvalidError{Expected: Mode{{ $mode }}, Actual: o.Mode}
+                return
+            }
+        {{- end }}
+
+        pairs = append(pairs, s.defaultPairs.{{ $fnk }}...)
```
Xuanwo, line 314: Is there any reason to move the logic here?

Prnyself: We should check and make sure `Object` is fit, otherwise we do not neet to handle pairs any more.

Xuanwo: LGTM

Xuanwo (approved):

PR #589, opened 2021-06-01
services: Implement GSP-90: Re-support Initialization Via Connection String

xxchan:
closes https://github.com/beyondstorage/specs/issues/56


cmd/definitions/tests/parse_test.go
```diff
@@ -0,0 +1,89 @@
...
+				assert.Nil(t, err)
+				assert.True(t, ok)
+			} else {
+				assert.True(t, errors.Is(err, tt.err))
+				return
+			}
+
+			sort.Sort(byPairName(service.Pairs))
+			sort.Sort(byPairName(tt.pairs))
+			assert.Equal(t, tt.pairs, service.Pairs)
```
Xuanwo, line 86: We can use `assert.ElementsMatch`

services/parse_pairs.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	. "github.com/beyondstorage/go-storage/v4/types"
+)
+
+var (
+	// ErrPairTypeNotParsable means the pair's type is not parseable.
+	ErrPairTypeNotParsable = NewErrorCode("type not parseable")
```
Xuanwo, line 14: Maybe we can remove those pairs from the map.

services/parse_pairs.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if err != nil {
+		pair = Pair{}
+		err = fmt.Errorf("%w: %v, %v, %v: %v", ErrPairValueInvalid, k, vType, v, err)
+	}
+	return
+}
+
+func parseListMode(s string) (ListMode, error) {
```
Xuanwo, line 76: Let's implement `FromString` for all those go-storage types.

services/parse_pairs.go
```diff
@@ -0,0 +1,102 @@
...
+	v64, err := strconv.ParseUint(s, 0, 8)
```
Xuanwo, line 77: the string is the human-readable representation, we should not allow user input list mode like `"2"` or `"0b11"`, just `"dir"` or `prefix|part`

services/parse_pairs.go
```diff
@@ -0,0 +1,145 @@
...
+		pair.Value, err = strconv.ParseBool(v)
+	case "int":
+		var i int64
+		i, err = strconv.ParseInt(v, 0, 0)
+		pair.Value = int(i)
+	case "int64":
+		pair.Value, err = strconv.ParseInt(v, 0, 64)
+	case "[]byte":
+		pair.Value, err = base64.RawStdEncoding.DecodeString(v)
+	case "ListMode":
```
Xuanwo, line 110: Maybe we can leave `ListMode` support in another GSP?

services/new.go
```diff
@@ -67,3 +67,34 @@ func init() {
...
+
+var (
+	servicePairMaps map[string]map[string]string
+)
+
+// RegisterSchema will register a service's pair map.
+//
+// Users SHOULD NOT call this function.
+func RegisterSchema(ty string, m map[string]string) {
+	servicePairMaps[ty] = m
```
Xuanwo, line 79: How about adding a lock to avoid race conditions?

services/parse_pairs.go
```diff
@@ -0,0 +1,117 @@
+package services
```
Xuanwo, line 1: How about move those code into `new`?

services/parse_pairs.go
```diff
@@ -0,0 +1,117 @@
...
+	. "github.com/beyondstorage/go-storage/v4/types"
+)
+
+var (
+	// ErrConnectionStringInvalid means the connection string is invalid.
+	ErrConnectionStringInvalid = NewErrorCode("connection string is invalid")
+)
+
+// <type>://[<name>][<work_dir>][?key1=value1&...&keyN=valueN]
+func parseString(ConnStr string) (ty string, ps []Pair, err error) {
```
Xuanwo, line 19: `parseString` looks too general to me, how about using `parseConnectionString`?

services/parse_pairs.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if len(path) == 0 {
+		// both <name> and <work_dir> missing
+	} else {
+		slash := strings.Index(path, "/")
+		if slash == -1 {
+			name := path
+			ps = append(ps, pairs.WithName(name))
+		} else if slash == 0 {
+			work_dir := path
```
Xuanwo, line 60: How about using `workDir` instead?

services/parse_pairs.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	for _, v := range strings.Split(rest, "&") {
+		opt := strings.SplitN(v, "=", 2)
+		if len(opt) != 2 {
+			// && or &key&, ignore
+			continue
+		}
+		pair, err1 := parse(m, opt[0], opt[1])
```
Xuanwo, line 75: `err1` looks wired to me, how about using `parseErr` or `return nil, err`

services/parse_pairs.go
xxchan: I agree this is not a good name, but does `ConnectionString` seem too long (I'm ok with it though)?

Xuanwo: Long but clear, I think it's OK.

Xuanwo (approved):

PR #590, opened 2021-06-03
pairs: Implement GSP-90: Parse Pairs from String (alternative)

xxchan:
alternative of https://github.com/beyondstorage/go-storage/pull/589

compare: https://github.com/beyondstorage/go-storage/compare/parse-pairs...parse-pairs-2

Using a deserializer ([gorilla/schema](https://github.com/gorilla/schema)) instead of implementing parse functions ourselves.

We will need a deserialization target (it includes the zero value of the actually interested type `T`) (can this be called a schema?). We can generate code to get such a target. `gorilla/schema` deserializes `url.Values`, so we generate a `&struct{ Value T }{} }`

pros:
- We don't need to hand-write parse functions for all types.
- Suppress the use of `reflect.Type`. Using generated typed values (put in `interface{}`)
   I think the reason why #589 needs `reflect` but this not is that the deserializer hides it for us.

cons:
- We can't control *whether* and *how* to parse a type?
- The reliability and safety of the deserializer library is unsure?

### alternative of `gorilla/schema`
With a more general deserializer abstraction (e.g., [Xuanwo/serde-go](https://github.com/Xuanwo/serde-go)), can we
- flexibly control how to parse different types?
- directly deserialize into `T`, without wrapped into a `struct`?


xxchan:
Will this style be better:

Let `New` returns directly an `interface{}` representing a new zero value for the pair's value type.

i.e., generate 
`PairInfo{New: func() interface{} { var v string; return v }}` 
instead of 
`PairInfo{New: func() interface{} { return &struct{ Value string }{} }},`

and add utilities 
```go
func toDeserializeTarget(value interface{}) interface{} {
	return reflect.New(reflect.StructOf([]reflect.StructField{
		{Name: "Value", Type: reflect.TypeOf(value)},
	})).Interface()
}

func fromDeserializeTarget(target interface{}) interface{} {
	return reflect.ValueOf(target).Elem().FieldByName("Value").Interface()
}


func parse() {
	// ...
	target := toDeserializeTarget(pairInfo.New())
	pair.Value = fromDeserializeTarget(target)
	// ...
}
```

Slightly easier to understand and modify target format?

xxchan:
benchmark

```go

func BenchmarkParseMap(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		pairs.ParseMap(Type, map[string]string{
			"name":          "abc",
			"storage_class": "sc",
			"size":          "12345",
		})

	}
}

```

This PR
``` go
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkParseMap-12    	  206767	      5656 ns/op	    1264 B/op	      50 allocs/op
```
Above comment
```
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkParseMap-12    	  172959	      7975 ns/op	    1519 B/op	      44 allocs/op
```

#589
```
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkParseMap-12    	 1000000	      1037 ns/op	     288 B/op	       7 allocs/op
PASS
```

PR #598, opened 2021-06-15
cmd/definitions: Trim Service Pair Prefix

xxchan:
Closes https://github.com/beyondstorage/specs/issues/98

Proposal https://github.com/beyondstorage/specs/pull/105

- Trim the type prefix of the internal names of service pairs.
- If a service pair has the same name as a global pair 
  - If their types are the same, a warning is logged.
  - If their types are different, generation fails.


cmd/definitions/type.go
```diff
@@ -115,10 +118,7 @@ func (p *Pair) Format(s specs.Pair, global bool) {
 
 // FullName will print full name for current pair
 func (p *Pair) FullName() string {
-	if p.Global {
-		return fmt.Sprintf("\"%s\"", p.Name)
-	}
-	return "pair" + templateutils.ToPascal(p.Name)
+	return fmt.Sprintf("\"%s\"", p.Name)
```
Xuanwo, line 121: Maybe we can remove `FullName` if there is no need for now?

Xuanwo (approved):

Issue #604, opened 2021-06-22
Tracking issue for GSP-117: Rename Service to System as the Opposite to Global


Xuanwo:
- [RFC PR](https://github.com/beyondstorage/specs/pull/117)
- [Rendered RFC](https://github.com/beyondstorage/specs/blob/master/rfcs/117-rename-service-to-system-as-the-opposite-to-global.md)

## Steps

- [x] Add `system metadata` for object meta in specs and go-storage https://github.com/beyondstorage/specs/pull/123
- [x] Mark `service metadata` related as deprecated https://github.com/beyondstorage/go-storage/pull/608
- [x] Update all references to `Object.serviceMetadata` in services
- [x] Update comments about `Service Pair`, `Service Info` and so on in site https://github.com/beyondstorage/site/pull/124

Xuanwo:
We will remove the deprecated fields in v4.3.0

PR #608, opened 2021-06-23
Implement GSP-117 Rename Service to System as the Opposite to Global

JinnyYi:
rfc: https://github.com/beyondstorage/go-storage/issues/604


cmd/definitions/tmpl/object.tmpl
```diff
@@ -52,7 +52,9 @@ func (o *Object) Set{{ $v.DisplayName }}(v {{ $v.Type }}) *Object {
 {{- else }}
 
 {{ if $isDeprecate }}
-// Deprecated: Moved to storage meta.
+// Get{{ $v.DisplayName }} will get {{ $v.DisplayName }} from Object.
```
Xuanwo, line 55: Maybe we don't need to check `{{ if $isDeprecate }}` anymore?

Xuanwo (approved):

PR #614, opened 2021-07-05
types: Add comments for Object and ObjectMode

Xuanwo:


types/object.go
```diff
@@ -5,6 +5,26 @@ import (
 	"sync/atomic"
 )
 
+// ObjectMode describes what users can on on this object.
```
xxchan, line 8: typo: on on

cmd/definitions/tmpl/object.tmpl
```diff
@@ -20,6 +20,8 @@ const (
 //   - Object's fields SHOULD not be changed outside services.
 //   - Object CANNOT be copied
 //   - Object is concurrent safe.
+//   - Only `ID`, `Path`, `Mode` are required during list operations, other fields
+//     could be fetched via lazy stat logic.
```
xxchan, line 24: What about linking the doc https://beyondstorage.io/docs/go-storage/internal/object-lazy-stat

types/object.go
```diff
@@ -5,6 +5,26 @@ import (
 	"sync/atomic"
 )
 
+// ObjectMode describes what users can operate on this object.
+//
+// Different object mode is orthogonal, an object could have different object mode
```
xxchan, line 10: Different object mode**s are** orthogonal, **and** an object could have different object mode**s** at the same time.

types/object.go
```diff
@@ -5,6 +5,26 @@ import (
...
+//
+// For example:
+//
+// - ModeDir means we can do list on this object('s path)
+// - ModeRead means we can read it as a normal file
+// - ModeLink means we can use we can read this object's target
+//
+// And we can compose them together:
+//
+// - ModeRead & ModeLink: Think about a symlink, we can still read it.
```
xxchan, line 21: Should we use `|` instead of `&`?

types/object.go
```diff
@@ -5,6 +5,26 @@ import (
...
+// - ModeDir & ModeLink: Think about a symlink to dir.
+// - ModeAppend & ModeRead: file in fs could be both read and append.
+//
+// Reference
+//
+// - [GSP-25](https://github.com/beyondstorage/specs/blob/master/rfcs/25-object-mode.md)
```
xxchan, line 27: What about linking the doc instead of the RFC (and maybe link GSP in the doc)? I think RFCs are not for users to read?

https://beyondstorage.io/docs/go-storage/internal/core-concept#object

types/object.go
```diff
@@ -5,6 +5,27 @@ import (
...
+// ObjectMode describes what users can operate on this object.
+//
+// Different object modes are orthogonal, and an object could have different
+// object modes at the same time.
+//
+// For example:
+//
+// - ModeDir means we can do list on this object('s path)
+// - ModeRead means we can read it as a normal file
+// - ModeLink means we can use we can read this object's target
```
JinnyYi, line 17: "we can use we can read" looks like a typo?

types/object.go
```diff
@@ -5,6 +5,27 @@ import (
...
+// And we can compose them together:
+//
+// - ModeRead | ModeLink: Think about a symlink, we can still read it.
+// - ModeDir | ModeLink: Think about a symlink to dir.
+// - ModeAppend | ModeRead: file in fs could be both read and append.
+//
+// Reference
+//
+// - [GSP-25](https://github.com/beyondstorage/specs/blob/master/rfcs/25-object-mode.md)
+// - [Core Concept: Object](https://beyondstorage.io/docs/go-storage/internal/core-concept#object)
```
xxchan, line 28: btw, go doc is not rendered in markdown. Do we need something like a docstring style guide?

xxchan (approved):

types/object.go
Xuanwo: Yes, we need a docstring style guide, maybe at the org level. I add an issue https://github.com/beyondstorage/site/issues/139 to track.

PR #615, opened 2021-07-05
definitions: Add support for type uint64

zu1k:
https://github.com/beyondstorage/go-service-ipfs/pull/3#commitcomment-53044856

### DIFF
--- a/cmd/definitions/parse.go
+++ b/cmd/definitions/parse.go
@@ -1,3 +1,4 @@
+//go:build tools
 // +build tools
 
 package main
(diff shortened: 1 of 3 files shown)


Xuanwo:
Looks like we need a `go fmt` here?

Xuanwo:
Nice!

zu1k:
My golang automatically adds a line: `//go:build tools`.

And it looks like there are a lot of files that need to be formatted.

Xuanwo:
> And it looks like there are a lot of files that need to be formatted.

We will support go 1.17 after it officially released. So doesn't matter, running `make build` to make sure all files formatted is OK to me.

xxchan:
Can you also add it here https://github.com/beyondstorage/go-storage/blob/master/services/new.go#L197-L202

Xuanwo (approved):

PR #628, opened 2021-07-09
docs: Move rfcs and spec here as required in GSP-139

Xuanwo:


Xuanwo:
All rfcs and spec are copied here, we only need to care about `docs/rfcs/README.md`.

JinnyYi (approved):

PR #652, opened 2021-07-09
cmd: Move definitions here as described in GSP-139

Xuanwo:
Maybe we can refactor them later, but in this PR, we just copy them here.


cmd/definitions/type.go
```diff
@@ -120,7 +121,7 @@ type Pair struct {
 }
 
 func (p *Pair) Type() string {
-	return parseType(p.ptype)
+	return p.ptype
```
xxchan, line 124: Why this is changed?

Xuanwo: Because we won't share definitions across languages, so we can remove the extra type translation layer here.

xxchan (approved):

Issue #525, opened 2021-04-16
Add generate diff check to make sure generated.go not changed


Xuanwo:
## Background

`go-storage` has build a codegen `definitions` (cmd/definitions) to help generated code for our services. And services will call the codegen in `doc.go` like:

https://github.com/beyondstorage/go-service-s3/blob/4217898092e96b6c2163f4281133eebd8856d90b/doc.go#L4-L6

```go
package s3

//go:generate go run -tags tools github.com/beyondstorage/go-storage/v4/cmd/definitions service.toml
```

But new contributors could miss the code generate and send PRs with not generated code. So It's better to add a check to make sure our `generated.go` generated correctly.

Moreover, this workflow checks not only `geneated.go` but also `go.mod` / `go.sum` and so on.

## Action

We can add a new workflow called `diff_check.yml` in which we:

- setup-go
- checkout services
- run `make build`
- run `git diff` and make sure the output is empty (just we did for `gofmt`)

burntcarrot:
Hmm, not sure about how to add it. 🤔
Would love a description so I can get a better understanding! 😀

Xuanwo:
> Hmm, not sure about how to add it. thinking
> Would love a description so I can get a better understanding! grinning

Updated, please take a look.

burntcarrot:
I have two options:

#### 1. Run job for each service

```
name: "diff_check"
on: [ push, pull_request ]

jobs:
  go-service-fs:
    name: go-service-fs
    runs-on: ubuntu-latest

    steps:
      - name: Set up Go 1.x
        uses: actions/setup-go@v2
        with:
          go-version: "1.15"

      - name: Checkout go-service-fs repo
        uses: actions/checkout@v2
        with:
          repository: beyondstorage/go-service-fs
          path: go-service-fs

      - name: Build
        run: make build

      - name: Git Diff
        run: if [ "$(git diff | wc -l)" -gt 0 ]; then exit 1; fi
  
  go-service-gcs:
    name: go-service-gcs
    runs-on: ubuntu-latest

    steps:
      - name: Set up Go 1.x
        uses: actions/setup-go@v2
        with:
          go-version: "1.15"

      - name: Checkout go-service-gcs repo
        uses: actions/checkout@v2
        with:
          repository: beyondstorage/go-service-gcs
          path: go-service-gcs

      - name: Build
        run: make build

      - name: Git Diff
[...]

PR #664, opened 2021-07-14
ci: add diff_check workflow

burntcarrot:
Added a generate diff check. Related #525.

### DIFF
--- a/.github/workflows/diff_check.yml
+++ b/.github/workflows/diff_check.yml
@@ -1,0 +1,246 @@
+name: "diff_check"
+on: [ push, pull_request ]
+
+jobs:
+  go-service-azblob:
+    name: go-service-azblob
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.15"
+
+      - name: Checkout go-service-azblob repo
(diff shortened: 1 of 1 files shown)


burntcarrot:
My guess is that I've not added the relative paths to each Makefile for each service, since you have to be in a directory which contains a Makefile. 

For example, the paths should be pointing to their respective services like `/home/runner/work/go-storage/go-storage/go-service-azblob` and not `/home/runner/work/go-storage/go-storage`.

Xuanwo:
We need to add a `diff_check.yml` under https://github.com/beyondstorage/go-service-s3/tree/master/.github/workflows like what we do for `gofmt.yml`. So we can run the check every time there is a new push or PR for `go-service-s3`.

burntcarrot:
> Sorry, I think we had some communication problems. I'm sorry I didn't read your example carefully in #525. We need to add `diff_check` for every single service instead of in `go-storage`. (So that we can check the diff while there are new push / PRs for services)

I was suspecting this earlier. Since `generated.go` lies in the repos of each service, hence, the actions should be run there.

Anyways, thanks for clarifying now! I'll make PRs for each service.

Xuanwo:
Closed as discussed in https://github.com/beyondstorage/go-storage/pull/664#issuecomment-879692592

PR #687, opened 2021-07-22
tests: Implement bench for List

Xuanwo:

### DIFF
--- a/tests/storager_bench_test.go
+++ b/tests/storager_bench_test.go
@@ -20,7 +20,7 @@ func BenchmarkStorage_Stat(b *testing.B) {
 
 func BenchmarkStorage_List(b *testing.B) {
 	ctx := context.TODO()
-	
+
 	var ob []*types.Object
 	for i := 0; i < 1024; i++ {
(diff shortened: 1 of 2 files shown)


tests/storager_bench_test.go
```diff
@@ -12,3 +16,23 @@ func BenchmarkStorage_Stat(b *testing.B) {
...
+		objects: ob,
+	}
+
+	it, err := s.list(nil, "", pairStorageList{})
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	for i := 0; i < b.N; i++ {
+		_, _ = it.Next()
```
JinnyYi, line 36: I'm confused about the target operation. Why is `Next` but not the complete list operation(retrieve all objects)?

Xuanwo: > I'm confused about the target operation. Why is `Next` but not the complete list operation(retrieve all objects)?

Because I want to measure the cost per `Next` operation.

JinnyYi (approved):

PR #690, opened 2021-07-23
CI: Add golangci-lint

burntcarrot:
Signed-off-by: burntcarrot <aadhav.n1@gmail.com>

Resolves #689.

PR #691, opened 2021-07-24
*: Fix ineffectual assignments

burntcarrot:
Related #690.


burntcarrot:
The errors returned by `structcheck` are I guess false positives, as these fields are referenced by code below it.

burntcarrot:
Added `nolint` to these fields, to ensure that the CI passes. Can you review the PR now?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,37 @@
+name: golangci-lint
```
Xuanwo, line 1: Please remove this action from this PR, as we will add it in #690

Xuanwo:
Others LGTM

Issue #680, opened 2021-07-18
Connection string needs to support config StorageFeatures and DefaultPairs


Xuanwo:

xxchan:
a possible format:
- `storage_features=loose_pair, virtual_dir`
- `default_storage_pairs.write=[key1=value&key2=value]`

Implementation problems
- The syntax: 
  - We used `&` to split pairs, and disallow `&` in key and value. If `&` is used, we should rewrite parser. 
  - The key were pair names. If we want to use the dot syntax, we should rewrite parser.
- `*Features` & `Default*Pairs` are types defined in services. We cannot use them as the parsed value type. So `go-service-*` should register either a parse function or `reflect.Type`(?)
- `default_storage_pairs` needs the ability to parse pairs. So this cannot be parsed within the service, we should either:
  - register `reflect.Type`
  - register a parse function which accepts another parse function
  - export pair parsing API

Xuanwo:
@JinnyYi, can you help continue the design?

JinnyYi:
> @JinnyYi, can you help continue the design?

OK, I'll give feedback in time if there's any progress or questions.

JinnyYi:
> * default_storage_pairs.write=[key1=value&key2=value]
> * A possible other compatible format is like `default_storage_pairs={write:[key1=value,key2=value],read:[]}` (a little bit ugly)

I'm not sure about the format for `default*pairs`. 
How about `default_storage_pairs.write=[key1=value, key2=value]&default_storage_pairs.write_append=[key1=value, key2=value]`? 
 It's similar to the first one. But they don't fit "the key is the pair name defined in toml and the format SHOULD be exactly the same" described in GSP-90?

Xuanwo:
> But they don't fit "the key is the pair name defined in toml and the format SHOULD be exactly the same" described in GSP-90?

It's OK to loosen the requirements.

> How about `default_storage_pairs.write=[key1=value, key2=value]&default_storage_pairs.write_append=[key1=value, key2=value]`?

I prefer `default_storage_pairs={write:[storage_class=STANDARD_IA,key2=value],read:[]}`

---

`default_storage_pairs` exposes too many internal details to users.

Maybe we don't need to support let users set up default pairs via connection strings. Instead, we can add new pairs like:

- default_storage_class => convert to write/write_append/wirte_pair storage class pair.
- default_sse_key => convert to read/write/write_append sse related keys.

[...]

PR #699, opened 2021-07-30
feat: Implement GSP-700 Config Features and DefaultPairs via Connection String

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/680


Xuanwo:
Please start RFC first before implement this feature.

tests/connstr_test.go
```diff
@@ -84,6 +84,30 @@ func TestFromString(t *testing.T) {
...
+			"tests://abc/tmp/tmp1?default_content_type=application/text&storage_class=sc",
+			[]Pair{
+				pairs.WithName("abc"),
+				pairs.WithWorkDir("/tmp/tmp1"),
+				WithDefaultContentType("application/text"),
+				WithStorageClass("sc"),
+			},
+			nil,
+		},
+		//{
```
Xuanwo, line 98: Do we need this test?

tests/generated.go
```diff
@@ -591,8 +681,36 @@ func parsePairStorageNew(opts []Pair) (pairStorageNew, error) {
 			}
 			result.HasWorkDir = true
 			result.WorkDir = v.Value.(string)
+		// Default pairs
+		case "default_content_type":
+			defaultPairs = true
```
Xuanwo, line 686: Maybe we can simplify the logic via respect to the order of pairs?

We don't need to take `default_storage_pairs` first any more, instead we always take the order.

- If user input `WithDefaultStoragePairs` and `WithDefaultStorageClass`, we will take `WithDefaultStoragePairs` first.
- If user input `WithDefaultStorageClass` and `WithDefaultStoragePairs`, we will take `WithDefaultStorageClass`.

After this change, we don't need to maintain `defaultPairs` or `enableFeatures`, we can set `HasStorageFeatures` directly.


NOTICE: this change requires an update of GSP.

JinnyYi: Does this mean that we do not consider the case where there is no intersection，like: 
```
oss.WithDefaultStoragePairs("Standard")
// ...
oss.WithDefaultStoragePairs(oss.DefaultStoragePairs{Write: []types.Pair{ps.WithContentType("application/text")}})
```

Xuanwo: Yes, we can use them concurrently, they will not affect each other.

JinnyYi: Maybe we need to keep these two variables `defaultPairs` and `enableFeatures` as when there is more than one default pairs(`DefaultStorageClass` and `DefaultContentType`, or `EnableVirtualDir` and `EnableVirtualLink`), we can't set `HasDefault*Pairs` or `Has*Features` to true every time, otherwise we would ignore some default values？

Xuanwo: `DefaultStorageClass` and `DefaultContentType` will always be appened into `DefaultStorageParis`, I think setting `HasDefaultStorageParis` is the same with `defaultPairs`.

cmd/definitions/tmpl/service.tmpl
```diff
@@ -103,7 +103,7 @@ func setStorageSystemMetadata(s *StorageMeta, sm StorageSystemMetadata) {
 }
 
 {{- range $_, $v := .Pairs }}
-{{- if not $v.Global }}
+{{- if and (not $v.Global) (not (hasPrefix $v.Name "enable_")) }}
```
JinnyYi, line 106: I removed `WithEnableXxx()` for features as we can use `<type>.WithStorageFeatures(<type>.StorageFeatures{Xxx: true, })` during initialize. I'm not sure whether it's necessary.

Xuanwo: I think it's OK to allow users to enable a feature via `WithEableXxx()`.

tests/generated.go
```diff
@@ -683,34 +663,21 @@ func parsePairStorageNew(opts []Pair) (pairStorageNew, error) {
...
+			result.HasDefaultStoragePairs = true
 			result.DefaultStoragePairs.Write = append(result.DefaultStoragePairs.Write, WithStorageClass(v.Value.(string)))
 		// Enable features
 		case "enable_loose_pair":
-			if result.HasStorageFeatures {
-				continue
-			}
-			enableFeatures = true
+			result.HasStorageFeatures = true
 			result.StorageFeatures.LoosePair = v.Value.(bool)
```
JinnyYi, line 675: I found the last one will be picked if there are multiple `enable_loose_pair`.

cmd/definitions/tmpl/service.tmpl
JinnyYi: OK, I'll roll back this change.

tests/generated.go
Xuanwo: How about generating `HasEnableLoose` to prevent it?

Maybe treat `enable_loose` as a normal pair can prevent those strange edge cases?

services/new.go
```diff
@@ -167,8 +167,12 @@ func parseConnectionString(ConnStr string) (ty string, ps []types.Pair, err erro
...
 		opt := strings.SplitN(v, "=", 2)
 		if len(opt) != 2 {
-			// && or &key&, ignore
-			continue
+			if v != "" {
+				opt = append(opt, v, "true")
+			} else {
+				// && , ignore
+				continue
+			}
```
JinnyYi, line 175: It's hard-coded here.
How about not generate pairs for `enable_xxx` features and register parsing functions for `StorageFeatures` and `ServiceFeatures`? Then we can collection pairs without values for `StorageFeatures` or `ServiceFeatures` parsing?

Xuanwo: Current implementation LGTM, we will refactor this part later, so don't be worried.

cmd/definitions/tmpl/service.tmpl
```diff
@@ -6,6 +6,20 @@ import (
...
+    {{- range $_, $dp := $v.Defaultable }}
+    {{- range $_, $sp := $pairs }}
+    {{- if and (eq $dp.Name $sp.Name) $sp.Global }}
+        {{- $needPair = true }}
+    {{- end }}
+    {{- end }}
+    {{- end }}
+    {{- end }}
+    {{- if $needPair }}
+    "github.com/beyondstorage/go-storage/v4/pairs"
```
Xuanwo, line 21: How about always import pairs? Like:

```
import "github.com/beyondstorage/go-storage/v4/pairs"

var _ pairs.XXXX
```

services/new.go
JinnyYi: Got it.

cmd/definitions/tmpl/service.tmpl
Xuanwo: For bool types, maybe we can generate `WithEnableXxxx()` directly? With the internal value set to `true`.

Xuanwo (commented):
I have some questions and let's discuss in https://matrix.to/#/#beyondstorage@gsp-700:matrix.org

Xuanwo:
The PR is too big, I updated steps in https://github.com/beyondstorage/go-storage/issues/704, let's split it into PRs.

Xuanwo:
Maybe we can close this PR for now?

PR #700, opened 2021-08-04
GSP: Config Features and DefaultPairs via Connetion String

JinnyYi:
previous discussion: https://github.com/beyondstorage/go-storage/issues/680
demo: https://github.com/beyondstorage/go-storage/pull/699

### DIFF
--- a/docs/rfcs/config-features-and-defaultpairs-via-connection-string.md
+++ b/docs/rfcs/config-features-and-defaultpairs-via-connection-string.md
@@ -1,0 +1,225 @@
+- Author: JinnyYi <github.com/JinnyYi>
+- Start Date: 2021-08-03
+- RFC PR: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
+
+# GSP: Config Features and DefaultPairs via Connection String
+
+Previous discussion:
+
+- [Connection string needs to support config StorageFeatures and DefaultPairs](https://github.com/beyondstorage/go-storage/issues/680)
+
+## Background
+
+In [GSP-90: Re-support Initialization Via Connection String](./90-re-support-initialization-via-connection-string.md), we introduced connection string to support service init and defined the format of connection string. 
+The connection string exposes basic types and not support complex types like `struct` at present. So we can't set `ServiceFeature`/`StorageFeatures` and `DefaultServicePairs`/`DefaultStoragePairs` by connection string during initialization.
+
+`Default*Pairs` is a slightly complex struct:
+
+```go
+// DefaultStoragePairs is default pairs for specific action
+type DefaultStoragePairs struct {
+	Create            []Pair
+	Delete            []Pair
+	List              []Pair
+	Metadata          []Pair
+	Read              []Pair
+	Stat              []Pair
+	Write             []Pair
+}
+```
+
+A possible format that follows the struct in connection string for default storage paris may be:
+
+ `default_storage_pairs={write:[storage_class=STANDARD_IA,key2=value],read:[]}`
+
+Obviously, It's complex for users and hard to pass in a valid string. In addition, it either exposes the internal details to end user directly (additional burden for end users) or it has to manipulate its config into our format (additional burden for application developers).
+
+## Proposal
+
+I propose to support config features and default pairs in plain string format in connection string.
+
+### Features
+
+The format of features pairs in connection string is:
+
+`enable_key=value`
+
+- The whole key of the pair SHOULD have the prefix `enable_`.
+- `key` is the pair name defined in `features.toml` and supported in service, and the format SHOULD be exactly the same.
+- `value` SHOULD be `true` or `false`.
+
+So a valid connection string containing features could be:
+
+- `s3://bucket_name/prefix?credential=hmac:xxxx:xxxx&enable_loose_pair=false&enable_virtual_dir=true`
+
+The connection string above is equivalent to:
+
+```go
+store, err := s3.NewStorage(
+	ps.WithCredential("hmac:xxxx:xxxx"),
+	ps.WithName("bucket_name"),
+	ps.WithWorkDir("/prefix"),
+	s3.WithStorageFeatures(s3.StorageFeaturs{LoosePair:  true,
+		LoosePair:  true,
+		VirtualDir: true,
+    })
+)
+```
+
+### DefaultPairs
+
+The format of default pair in connection string is:
+
+`default_key=value`
+
+- The whole key of the pair SHOULD have the prefix `default_`.
+- `key` is the pair name defined in `toml` and the format SHOULD be exactly the same.
+- Operation has a different set of default pairs. Pairs in operation pair list will share the same default value if have the same name .
+
+So a valid connection string containing default pairs could be:
+
+- `s3://bucket_name/prefix?credential=hmac:xxxx:xxxx&default_server_side_encryption=xxxx&default_strorage_class=xxxx`
+
+Also, we generate the following function for default pairs to support configure the default pair during initialization in the way of `WithXxx()`:
+
+Take `default_storage_class` for example:
+
+```go
+func WithDefaultStorageClass(v string) Pair {
+	reutrn Pair{
+		Key:   "default_storage_class",
+		Value: v,
+	}
+}
+```
+
+The connection string above is equivalent to:
+
+```go
+store, err := s3.NewStorage(
+	ps.WithCredential("hmac:xxxx:xxxx"),
+	ps.WithName("bucket_name"),
+	ps.WithWorkDir("/prefix"),
+    }), 
+    s3.WithDefaultServiceSideEncryption("xxxx"),
+    s3.WithDefaultStorageClass("xxxx"),
+)
+
+```
+
+### Implementation
+
+#### Features
+
+`*Features` are types defined in service.
+
+**Feature Pairs Registry**
+
+From services side:
+
+We generate pairs like `enable_virtual_dir` according to `features` in `service.toml`. And register the feature pairs referring to [GSP-90: Re-support Initialization Via Connection String].
+
+```go
+var serviceFeaturesPairMap = map[string]string{
+    // ...
+    "default_loose_pair": "bool",
+}
+
+var storageFeaturesPairMap = map[string]string{
+    // ...
+    "default_loose_pair":  "bool",
+    "default_virtual_dir": "bool",
+}
+
+func init() {
+    // insert service and storage feature maps into `pairMap` for registry
+	for k, v := range serviceFeaturesPairMap {
+        pairMap[k] = v
+    }
+    for k, v := range storageFeaturesPairMap {
+        pairMap[k] = v
+    }
+    services.RegisterSchema("<type>", pairMap)
+}
+
+```
+
+According to the current logic, `New*FromString` will first split connection string parse it into `ps []Pairs`, and finally call `New*(ty, ps)`. Then we can handle the default feature pairs to convert to `*Features` in `ParsePair*New()`.
+
+#### Default Pairs
+
+Default pairs are defined in services.
+
+From [go-storage] side:
+
+- Defaultable field can be added to `Pair` in `specs` to mark whether the pair could be set as default while initiate.
+- Defaultable string array can be added to `Op` in `specs` to store the list of defaultable pair's name.
+- Defaultable pair array can be added to `Functions` in `definitions` to store the list of defaultable pairs.
+
+Based on the above, we can record all the default pairs and the default pairs for a specific operation while parsing service toml.
+
+From services side:
+
+We can add new field `defaultable` for operations, in which service can specify pairs that can be set as default, like:
+
+```toml
+[namespace.storage.op.write]
+optional = ["content_md5", "content_type", "io_callback", "storage_class"]
+defaultable = ["storage_class"]
+
+[namespace.storage.op.create_append]
+optional = ["content_type", "storage_class"]
+defaultable = ["storage_class"]
+```
+
+**Default pairs registry**
+
+We generate pairs like `default_storage_class` according to the defaultable pairs in the service. Then insert the pairs into `pairMap` like what we did in [GSP-90: Re-support Initialization Via Connection String]:
+
+```go
+var pairMap[string]string {
+	// ...
+	"default_storage_class" "string",
+}
+```
+
+The defaultable pairs will be registered based on the existing logic.
+
+**Parse pairs in `parsePair*New`**
+
+We can generate the following struct to carry the shared default pairs:
+
+```go
+type DefaultConfigs struct {
+	HasDefaultStorageClass  bool
+	DefaultStorageClass     string
+	...
+}
+```
+
+Then we can handle the default pairs to assign `DefaultConfigs` in `ParsePair*New()`.
+
+**Parse pairs in specific operation**
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+
+### Implementation
+
+#### Features
+
+`*Features` are types defined in service.
+
+**Feature Pairs Registry**
```
Xuanwo, line 117: We already have features field in `service.toml`:

```toml
features = ["virtual_dir"]
```

Maybe we can reuse it?

Template that uses features can be found here: https://github.com/beyondstorage/go-storage/blob/master/cmd/definitions/tmpl/service.tmpl#L135-L146

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+[namespace.storage.op.write]
+optional = ["content_md5", "content_type", "io_callback", "storage_class"]
+defaultable = ["storage_class"]
+
+[namespace.storage.op.create_append]
+optional = ["content_type", "storage_class"]
+defaultable = ["storage_class"]
+```
+
+**Default pairs registry**
```
Xuanwo, line 176: The same question here.

How about generating logic in `generated.go`?

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+The defaultable pairs will be registered based on the existing logic.
+
+**Parse pairs in `parsePair*New`**
+
+We can generate the following struct to carry the shared default pairs:
+
+```go
+type DefaultConfigs struct {
```
Xuanwo, line 194: Maybe we can generate into `defaultPairs` directly?

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
JinnyYi: Does it mean we can insert `map["virtual_dir"] = "bool"` into `pairMap` instead of generate `*FeaturesPairMap`?

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
JinnyYi: Got it.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
JinnyYi: Got it. Then there's no need to add an extra field in `Storage` and `Service`.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: https://github.com/beyondstorage/go-storage/blob/master/cmd/definitions/type.go#L283

Maybe we can append into New functions `Optional` directly?

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+
+```go
+store, err := s3.NewStorage(
+	ps.WithCredential("hmac:xxxx:xxxx"),
+	ps.WithName("bucket_name"),
+	ps.WithWorkDir("/prefix"),
+	s3.WithStorageFeatures(s3.StorageFeaturs{LoosePair:  true,
+		LoosePair:  true,
+		VirtualDir: true,
+    })
```
xxchan, line 66:
```suggestion
	s3.WithStorageFeatures(s3.StorageFeaturs{
		LoosePair:  false,
		VirtualDir: true,
    })

```

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+Take `default_storage_class` for example:
+
+```go
+func WithDefaultStorageClass(v string) Pair {
+	reutrn Pair{
+		Key:   "default_storage_class",
+		Value: v,
+	}
+}
+```
```
xxchan, line 95: What's its relationship with the old `DefaultStoragePairs`?

JinnyYi: I think `DefaultStoragePairs` is more fine-grained compaired to `WithDefaultStorageClass`.
`DefaultStoragePairs` is an embeded struct contains a set of `[]Pair` for supported operations. So users need to assign value to pairs of each operation one by one, even when different operations have the same pair.
For `WithDefaultStorageClass`,  pairs that have the same name will share the same default value，users needn't to specify the operation.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
JinnyYi: Got it.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,225 @@
...
+)
+```
+
+### DefaultPairs
+
+The format of default pair in connection string is:
+
+`default_key=value`
+
+- The whole key of the pair SHOULD have the prefix `default_`.
```
JinnyYi, line 76: Maybe we don't need the prefix `default_`? Then we can reuse `pairMap` with no change and `WithStorageClass` during initialization instead of  generating `WithDefaultStrageClass`?

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
xxchan: The two kind of pairs can be used together (not orthogonal). Will it make users confused? Will they conflict with each other?

JinnyYi: There's no confilic. I found that there's no need to generate `WithDefaultXxx()` functions. Maybe we can just use the existing functions `WithXxx()` to set the default paris?

JinnyYi: > Will they conflict with each other?

When both `WithDefault*Pairs()` and `WithXxx()` are used, I think the values passed in by `WithDefault*Pairs()` should be picked (the priority is: `pairs from args` > `WithDefault*Pairs()` > `WithXxx`), and this can be and should be generated.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: Ok, I got it, we are discussing different things.

There are two things we need to support:

- `DefaultStorageClass` pairs
- `default_storage_class` in connection string.

For the first, we can add pairs into `Optional`, and for the other, we can add into `pairMap`.

Finally, we can reduce the usage of `Feature Pairs Registry` or `Default Pairs Registry`.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: > Maybe we can just use the existing functions `WithXxx()` to set the default paris?

I don't think so.

They have different semantic meanings.

- `WithStorageClass` set current operation's storage class
- `WithDefaultStorageClass` set current storage's default storage class.

So `WithDefaultStorageClass` is not a valid pair for `Write` operation, only valid for `NewStorage` operation. And `WithStorageClass` is valid for `Write` but not for `NewStorage`.

NOTES: there is no direct conversion between `WithDefaultStorageClass` and `WithStorageClass`. `WithDefaultStorageClass` configure the `DefaultPairs` and `DefaultPairs` will be expand in corresponding operations.

If `loose_pair` is enabled, they will be ignored, otherwise, we will raise not support pairs if used incorrectly.

And to the next question: How about using `WithDefaultPairs` and `WithDefaultStorageClass` at the same time?

We will pick `DefaultPairs` first, and append pairs that parsed from `DefaultStorageClass`.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: As described in https://github.com/beyondstorage/go-storage/pull/700#discussion_r683939730, I don't think so.

Xuanwo:
Let service specify `default_xxx` may be a huge burden. Think about `s3`, we have to write `default_storage_class`, `default_excepted_bucket_owner` and so on.

Maybe we can introduce new fields to help services simplify the work? I camp up to two ways:

## Add new fields in operations

Like we discussed in https://github.com/beyondstorage/go-storage/issues/680, we can add a new field `defaultable` in operations. Paris that is listed in `defaultable` will be treated as allowing users to set default values. We can parse the toml files while generating and get a full list of all defaultable pairs, and then, generate related pairs.

```toml
[namespace.storage.op.write]
optional = ["content_md5", "content_type", "io_callback", "storage_class", "excepted_bucket_owner", "server_side_encryption_bucket_key_enabled", "server_side_encryption_customer_algorithm", "server_side_encryption_customer_key", "server_side_encryption_aws_kms_key_id", "server_side_encryption_context", "server_side_encryption"]
defaultable = ["content_type", "storage_class", "excepted_bucket_owner", ...]
```

## Add new fields in the namespace

Add new fields in operations could still be too verbose. Think about the same pair in different operations:

```
[namespace.storage.op.read]
optional = ["offset", "io_callback", "size", "excepted_bucket_owner"]
defaultable = ["excepted_bucket_owner"]

[namespace.storage.op.write]
optional = ["content_md5", "content_type", "io_callback", "storage_class", "excepted_bucket_owner"]
defaultable = ["excepted_bucket_owner"]
```

So maybe we can add them in the namespace, alongside `features` and `implement`.

```toml
[namespace.storage]
features = ["virtual_dir"]
implement = ["direr", "multiparter"]
defaultable = ["excepted_bucket_owner"]

[namespace.storage.op.read]
optional = ["offset", "io_callback", "size", "excepted_bucket_owner"]

[namespace.storage.op.write]
optional = ["content_md5", "content_type", "io_callback", "storage_class", "excepted_bucket_owner"]
```

---

No matter which way we choose, services only need to specify once instead of list all pairs in the `New` function.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,178 @@
...
+- Add pairs for the optional pairs of `New` with `default_` and `enable_` prefix.
+- Mark them as defaultable, the same with global and system paris corresponding to the default ones. 
+
+**Parse features in `parsePair*New`**
+
+We can handle the optional defaultable pairs with prefix `enable_`, assign value to the corresponding field of `*Features` in `ParsePair*New()`.
+
+**Parse default pairs in `parsePair*New`**
+
+We can add private fields correspond to default pairs into the generated `Default*Pairs` to carry the shared values, like:
```
Xuanwo, line 140: Maybe we can convert `default_storage_class=xxxx`/`WithStoragClass("xxx")` to pairs in `DefaultStoragePairs` directly?

- We know storage_class is a defaultable pair
- We know which operations support storage_class

JinnyYi:
> Maybe we can introduce new fields to help services simplify the work

Yes, and introduce new fields can specify the defaultable pairs more explicitly and visually. 

For `add new fields in operations`, it's easy to check the validity of the defaultable pairs.  We can even specify the defaultable paris on the operation level, other than the namespace level(all operations affected). But we need to traverse the operations' pair list once to collect all the defaultable paris.

For `add new fields in the namespace`, we can get all the defaultable pairs easily. But we also need to traverse the operations' pair list to check the validity of the pairs on namespace level.

I’ll try to `add new fields in the namespace`.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -113,45 +113,31 @@ That means pair of operations with the same name `server_side_encryption` or `st
...
-- Check whether the default pair belongs to global or system pair and feature belongs to the supported ones.
-- Add pairs for the optional pairs of `New` with `default_` and `enable_` prefix.
-- Mark them as defaultable, the same with global and system paris corresponding to the default ones. 
+- Check whether the key in `defaultable` belongs to the pair list of operations in the namespace.
+- Generate default pairs according to `defaultable` and `features` fields.
 
 **Parse features in `parsePair*New`**
 
-We can handle the optional defaultable pairs with prefix `enable_`, assign value to the corresponding field of `*Features` in `ParsePair*New()`.
+We can handle the defaultable pairs for features with prefix `enable_`, assign value to the corresponding field of `*Features` in `ParsePair*New()`.
```
Xuanwo, line 136: features will have `WithEnableVirtualDir`?

JinnyYi: Maybe `WithEnableVirtualDir()` is not necessary？I intended to add default paris for features to reduce the usage of `Feature Pairs Registry` for connection string, so `WithEnableVirtualDir()` will be generated if no additional controls.

Xuanwo: What does `add default paris for features` mean?

JinnyYi: Like pairs for `defaultable pairs`. Names of these pairs are prefixed with `enable_`. Then the keys `enable_xxx` will be added into `pairMap` and can be parsed when using connection string.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
xxchan: Do you think it is necessary to explain the rationale why we need to add new pairs? Since the RFC title is about connection string.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: OK, mostly LGTM.

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
```diff
@@ -0,0 +1,179 @@
+- Author: JinnyYi <github.com/JinnyYi>
+- Start Date: 2021-08-03
+- RFC PR: [beyondstorage/go-storage#700](https://github.com/beyondstorage/go-storage/issues/700)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
```
Xuanwo, line 4:
```suggestion
- Tracking Issue: [beyondstorage/go-storage#704](https://github.com/beyondstorage/go-storage/issues/704)
```

docs/rfcs/700-config-features-and-defaultpairs-via-connection-string.md
Xuanwo: Nice idea.

Xuanwo:
I approve of this GSP mostly. Let's explain the rationale why we need to add new pairs as described in https://github.com/beyondstorage/go-storage/pull/700#discussion_r685806637

Xuanwo (approved):

Xuanwo:
Nice work!

Issue #704, opened 2021-08-10
Tracking issues of GSP-700: Config Features and DefaultPairs via Connetion String


Xuanwo:
- RFC PR: https://github.com/beyondstorage/go-storage/pull/700

## Steps

- [x] Generate feature pairs (Allow using `EnableVirtualDir()`)
- [x] Generate default pairs (Allow using `DefaultStorageClass`, as described in GSP)
- [x] Add support in connection string (All using `s3://xxxx?default_storage_class=xxxx`)
- [x] Auto set `StorageFeatures` with input like `EnableVirtualDir` 
- [x] Auto set `DefaultStoragePairs` with input like `DefaultStorageClass`
- [ ] Add docs in site
- [ ] Add examples

Xuanwo:
Great, all code-related work has been done!

Xuanwo:
This issue will be extended by https://github.com/beyondstorage/go-storage/issues/727

PR #708, opened 2021-08-17
feat: Generate feature pairs

JinnyYi:


cmd/definitions/tmpl/service.tmpl
```diff
@@ -98,6 +98,14 @@ func setStorageSystemMetadata(s *StorageMeta, sm StorageSystemMetadata) {
 // With{{ $pname }} will apply {{ $v.Name }} value to Options.
 //
 {{ $v.Description }}
+{{- if hasPrefix $v.Name "enable_" }}
```
Xuanwo, line 101: Maybe we can check the `bool` type instead?

Is it OK to generate `EnableXxx()` or `DisableXxxx()` for all services?

cmd/definitions/type.go
```diff
@@ -660,6 +661,18 @@ func mergePairs(global, service map[string]*Pair) map[string]*Pair {
 	return ans
 }
 
+func addFeaturePairs(pairs map[string]*Pair, ns specs.Namespace) {
```
Xuanwo, line 664: How about implementing them in `FormatNamespace` directly?

cmd/definitions/type.go
```diff
@@ -660,6 +661,18 @@ func mergePairs(global, service map[string]*Pair) map[string]*Pair {
...
+	for _, feature := range ns.Features {
+		featurePair := &Pair{
+			Name:        "enable_" + feature,
```
Xuanwo, line 667: I prefer `name := fmt.Sprintf("enable_%s", feature)` and use `name` in the following.

cmd/definitions/type.go
```diff
@@ -660,6 +661,18 @@ func mergePairs(global, service map[string]*Pair) map[string]*Pair {
...
+			ptype:       "bool",
+			Global:      false,
+			Description: formatDescription(templateutils.ToPascal("enable_"+feature), "enable "+feature+" features."),
```
Xuanwo, line 670: Can we read the description of this feature?

cmd/definitions/tmpl/service.tmpl
JinnyYi: There's `disable_uri_cleaning` pair in our test toml. `EnableDisableURICleaning` or `DisableDisableURICleaning` looks strange. Take this scenario into consideration, how about consider only features for now and modify the condition without using prefix？

Xuanwo: Well, here is some misunderstanding.

- features will always be generated into `EnableXxxx` as described in GSP.
- Difference here: `disable_uri_cleaning` will be generated into `DisableURICleaning()` instead of `DisableURICleaning(value bool)`

`disable_uri_cleaning` here is a pair instead feature, so we will not add the prefix `enable_` for it.

JinnyYi: Got it.

Xuanwo:
What a simple and elegant implementation :+1: 

Nick work!

Xuanwo (approved):

PR #710, opened 2021-08-18
feat: Generate default pairs

JinnyYi:


cmd/definitions/type.go
```diff
@@ -74,15 +74,32 @@ func (s *Service) Pairs() []*Pair {
...
 }
 
+// Defaultable returns sorted defaultable pairs.
+func (n *Namespace) Defaultable() []*Pair {
+	keys := make([]string, 0, len(n.defaultable))
+
+	for k := range n.defaultable {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
Xuanwo, line 94: How about using `sort.Slice` on `[]*Pair` instead?

cmd/definitions/type.go
```diff
@@ -593,6 +628,35 @@ func (d *Data) ValidateNamespace(srv *Service, n *Namespace) {
 			log.Fatalf("Operation [%s] requires Pair [%s] support, please add virtual implementation for this pair.", v.Name, ps)
 		}
 	}
+
+	// Check if defaultable pairs are included in functions pairs.
+	for defaultablePairName := range n.defaultable {
```
Xuanwo, line 633: Can you explain a bit more about this validate logic?

JinnyYi: The intention is to check whether the default pairs belongs to the pairs of functions under namespace. For example, if `location` belongs to `service.op.create`, and no function in `storage` has this pair, then we can't add `location` as the defaultable pair for `storage`.

Xuanwo: I think it's no so harmful that we need to check so strictly. How about adding a new issue to record this problem and remove the check logic in this PR.

From my side, I think the best way is we maintain a pairs map in `namespace` and we only need to check is this pair exist in it. But we only have `srv.pairs` now.

JinnyYi: Got it.

Xuanwo (approved):

Xuanwo:
So nice!

PR #718, opened 2021-08-24
feat: Allow append pairs for  New{Servicer,Storager}FromString

Xuanwo:

### DIFF
--- a/services/new.go
+++ b/services/new.go
@@ -89,20 +89,22 @@ func RegisterSchema(ty string, m map[string]string) {
 	servicePairMaps[ty] = m
 }
 
-func NewServicerFromString(connStr string) (types.Servicer, error) {
-	ty, ps, err := parseConnectionString(connStr)
+func NewServicerFromString(connStr string, ps ...types.Pair) (types.Servicer, error) {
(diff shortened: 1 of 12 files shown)


services/new.go
```diff
@@ -89,20 +89,26 @@ func RegisterSchema(ty string, m map[string]string) {
 	servicePairMaps[ty] = m
 }
 
-func NewServicerFromString(connStr string) (types.Servicer, error) {
-	ty, ps, err := parseConnectionString(connStr)
+// NewServicerFromString will create a new service via connection string.
+func NewServicerFromString(connStr string, ps ...types.Pair) (types.Servicer, error) {
+	ty, psc, err := parseConnectionString(connStr)
 	if err != nil {
 		return nil, InitError{Op: "new_servicer", Type: ty, Err: err, Pairs: ps}
```
JinnyYi, line 96: Is it necessary to set `psc` to `Pairs` here(but `pcs` will be nil if parse failed)?

Xuanwo: It's not necessary but also not harmful. I prefer to not change it until we plan to refactor the connection parse logic.

JinnyYi (approved):

PR #719, opened 2021-08-24
feat: Set DefaultPairs with defaultable pairs

JinnyYi:


cmd/definitions/tmpl/service.tmpl
```diff
@@ -245,6 +248,33 @@ func parsePair{{ $pn }}{{ $fnk }}(opts []Pair) (pair{{ $pn }}{{ $fnk }}, error)
     }
     {{- end }}
 
+    // Default pairs
+    {{- range $_, $pair := $dp }}
+    if result.has{{ $pair.Name | toPascal }} {
+        result.HasDefault{{ $pn }}Pairs = true
+        {{- range $_, $op := $fu }}
+            {{- range $_, $opPair := $op.Required }}
+                {{- if eq $pair.Name (print "default_" $opPair.Name) }}
```
Xuanwo, line 257: Why? I thought we have `default_xxx` pairs already.

JinnyYi: I'm not so sure about your query.
The intention here is to find operations that contains the default pair, and the pairs for operations don't contain `default_` prefix.

Xuanwo: I'm a bit confused, let's discuss in https://matrix.to/#/#beyondstorage@gsp-700:matrix.org

cmd/definitions/type.go
```diff
@@ -85,19 +85,34 @@ type Namespace struct {
...
-func (n *Namespace) Defaultable() []*Pair {
-	ps := make([]*Pair, 0, len(n.defaultable))
+// Defaultable returns defaultable pair and operations map.
+func (n *Namespace) Defaultable() map[*Pair][]string {
+	pf := make(map[*Pair][]string)
+
 	for _, v := range n.defaultable {
-		ps = append(ps, v)
+		var ops []string
+		for _, op := range n.Funcs {
```
Xuanwo, line 94: How about

```go
ps := make([]*Paris, 0)
ps = append(ps, op.Required...)
ps = append(ps, op.Optional...)

for _, v := range ps {
    ...
}
```

Xuanwo (approved):

Issue #607, opened 2021-06-23
Use code generator to replace template 


Xuanwo:
Maybe we can use https://github.com/dave/jennifer to replace our huge templates?

From

```go
package main

import (
    "fmt"

    . "github.com/dave/jennifer/jen"
)

func main() {
	f := NewFile("main")
	f.Func().Id("main").Params().Block(
		Qual("fmt", "Println").Call(Lit("Hello, world")),
	)
	fmt.Printf("%#v", f)
}
```

To

```go
package main

import "fmt"

func main() {
	fmt.Println("Hello, world")
}
```

Does it deserve it?

Xuanwo:
`jennifer` looks too low level for us.

PR #728, opened 2021-08-27
feat: Use gg to generate code for pairs

Xuanwo:
This is a demo PR for evaluations.

Related to https://github.com/beyondstorage/go-storage/issues/607.

- Is go-codegen more friendly than go template?
- Is go-codegen easier to learn/use/understanding?


cmd/definitions/generate.go
```diff
@@ -42,6 +43,63 @@ func generateGlobal(data *Data) {
...
+		fn.WithParameters(codegen.NewParameter("v", v.Type(), false))
+		fn.WithResults(codegen.NewResult("p", "Pair", false))
+		fn.WithBody(`
+return Pair{
+	Key: "{{ .Name }}",
+	Value: v,
+}
+`, v)
+
+		f.AddFunction(fn)
```
JinnyYi, line 86: It's more readable and easy to debug. But I wondered whether the whole source file will be complicated for more  complex function body and more complex types?

xxchan: What do you mean by "easy to debug"?

Xuanwo: > more complex function body and more complex types

It's indeed a question, I can't answer for now (:

JinnyYi: I have encountered errors during code generation when using templates before and could not visually locate the error in the template. I think maybe this approach will be easier in dealing with this type of problem？

Xuanwo:
ping @JinnyYi  to take another look.

Xuanwo:
I will extend it to more complex functions like operations  tomorrow.

Xuanwo:
I re-design the API of codegen that inspired by `jen`, please take a look.

The code can't be compiled so fat, I will implement it if it looks good.

JinnyYi:
> I re-design the API of codegen that inspired by `jen`, please take a look.

I think it‘s easy to read and translate into the final generated code.

Xuanwo:
ping @JinnyYi for review.

JinnyYi:
LGTM. Only a small doubt, is it possible or necessary to keep only `Description` for `Pair`?

Xuanwo:
> LGTM. Only a small doubt, is it possible or necessary to keep only `Description` for `Pair`?

Yes, we have to. We use `originalDescription` here because we need to make sure the remaining templates works. We will remove `originalDescription` finally!

Xuanwo:
I will bump a version about gg, please merge this PR after update, thanks!

JinnyYi (approved):
LGTM.

PR #729, opened 2021-08-27
GSP-729: Redesign HTTP Signer

JinnyYi:


docs/rfcs/729-redesign-http-signer.md
```diff
@@ -0,0 +1,169 @@
...
+
+```go
+type HTTPSigner interface {
+    QuerySignHTTPRead(path string, expire time.Duration, ps ...types.Pair) (signedReq *http.Request, err error)
+    QuerySignHTTPWrite(path string, size int64, expire time.Duration, ps ...types.Pair) (signedReq *http.Request, err error)
+    // ...
+}
+```
+
+- `Read` and `Write` (add `Multiparter` related operations if needed) are the supported signature operations for now. Other operations SHOULD be introduced by new GSP.
```
Xuanwo, line 35: So we will only a `HTTPSigner` instead of `MultipartHTTPSinger`, `AppendHTTPSigner` and so on?

docs/rfcs/729-redesign-http-signer.md
```diff
@@ -0,0 +1,169 @@
...
+- RFC PR: [beyondstorage/go-storage#729](https://github.com/beyondstorage/go-storage/issues/729)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
+
+# GSP-729: Redesign HTTP Signer
+
+- Updates:
+  - [GSP-706]: Deprecate `QuerySignHTTP()`.
+
+Previous discussion:
+- https://github.com/beyondstorage/go-storage/issues/726
```
Xuanwo, line 12:
```suggestion
- [How to pass `partIndex` to `QuerySignHTTP` for `WriteMultipart`?](https://github.com/beyondstorage/go-storage/issues/726)
```

docs/rfcs/729-redesign-http-signer.md
```diff
@@ -0,0 +1,169 @@
...
+- https://matrix.to/#/#beyondstorage@gsp-706:matrix.org
```
Xuanwo, line 13:
```suggestion
- [#beyondstorage@gsp-706:matrix.org](https://matrix.to/#/#beyondstorage@gsp-706:matrix.org)
```

docs/rfcs/729-redesign-http-signer.md
JinnyYi: Maybe we should group them by interfaces? And for `StoragerHTTPSigner`, I'm not sure whether we need to add operations for all the non-local functions or only for `read/write` temporarily (as some services only support add signatures to URLs that are contained in PUT and GET requests, or only for GET)?

Xuanwo: I prefer `StorageHTTPSigner`.

How about adding them one by one, and leave this question for the smarted us in the further to decide.

`Read/Write` is enough so far.

Xuanwo:
Mostly LGTM, let's create a tracking issue for it!

Xuanwo:
Let's rock!

Issue #727, opened 2021-08-26
Tracking issue of GSP-725: Add Defaultable Property for Pair


JinnyYi:
- [Rendered RFC](https://github.com/beyondstorage/go-storage/pull/725)

## Steps

- [x] Add `defaultable` field in pair (definitions, cmd/specs)
- [x] Remove `defaultable` field in namespace (cmd/specs)
- [x] Generate default pairs for defaultable global pairs and defaultable system pairs (cmd/definitions/type.go)

PR #730, opened 2021-08-30
cmd/definitions: Implement GSP-725 Add Defaultable Property for Pair

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/727


tests/service.toml
```diff
@@ -51,6 +51,7 @@ type = "bool"
 
 [pairs.storage_class]
 type = "string"
+defaultable = true
```
Xuanwo, line 54: It looks like `WithDefaultStorageClass` is not generated.

JinnyYi: It has been generated in https://github.com/beyondstorage/go-storage/pull/730/files#diff-d872ad7215a358bcc5ae66b3577ab2db501f7f8497ac6624e7374a19b7f69fedR88. We pointed it as a defaultable pair in namespace before, so we can't find it in this change.

Xuanwo (approved):

Issue #731, opened 2021-09-01
Tracking issue of GSP-729: Redesign HTTP Signer


JinnyYi:
- [Rendered RFC](https://github.com/beyondstorage/go-storage/pull/729)

## Steps

- [x] Remove `HTTPSigner` and `QuerySignHTTP()`.
- [x] Add `StorageHTTPSigner` interface and operations in definitions. https://github.com/beyondstorage/go-storage/pull/732
- [x] Implement integration test. https://github.com/beyondstorage/go-integration-test/pull/49
- [x] Update docs in site (Add descriptions for the operaiton and a doc for migration from `Reacher` to `QuerySignHTTPRead`). https://github.com/beyondstorage/site/pull/191
- [x] Add example https://github.com/beyondstorage/go-storage-example/pull/21 https://github.com/beyondstorage/go-storage-example/pull/22

Xuanwo:
Maybe we need to release v4.7.0 before we implement it service side?

PR #734, opened 2021-09-02
definitions: Deprecate Reacher and expire

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/731


Xuanwo (approved):

Xuanwo:
Maybe we also need to deprecate `expire`?

Issue #744, opened 2021-09-06
1st Cleanup Week Roadmap


Xuanwo:
As announced at https://forum.beyondstorage.io/t/1st-cleanup-week-for-go-storage/213, this week (09/06-09/10) is the first cleanup week for go-storage.

We expect to stop any new feature development of go-storage and focus on bugfix, docs like:

- every project's `README.md` update (update links of matrix room, integration tests, and so on)
- go-storage-example enhancement.
- go-storage user guide and developer guide.
- Clean up issues we are working on and plan to do.
- File new issues about any problems that meet (could be solved later)
- ...

**Please feel free to open a new request in this issue!**

## Docs

- [x] https://github.com/beyondstorage/go-storage/issues/743

## Tests

- [x] #745

## Infra

- [x] Auto merge upgrade of go-storage @Xuanwo 

## Behavior

- [ ] https://forum.beyondstorage.io/t/specify-the-behavior-for-writing-empty-file/204/2
- [ ] https://forum.beyondstorage.io/t/specify-the-behavior-of-path/195/2
- [ ] https://forum.beyondstorage.io/t/proposal-adopt-commitizen-conventional-commit-types/193/2

JinnyYi:
I'll clean up issues related with HTTP Signer, including docs and example. In addition, I'd like to take over `Bebavior`,  go-storage-example enhancement.

abyss-w:
I can take over the `Tests` part first. I can also take over the `Behavior` part(the path or write empty file).

Xuanwo:
Hi, it looks like we are working on the `Behavior` part now. Maybe we need to leave them as the last part? 

Here is some projects reply to the new releases of various services. How about finishing #743 and #745 first, and release new versions for all our services first?

JinnyYi:
> How about finishing #743 and #745 first, and release new versions for all our services first?

I can embark on  https://github.com/beyondstorage/go-storage/issues/743 refer to https://github.com/beyondstorage/go-service-s3/pull/208.

Xuanwo:
Nice work!

PR #749, opened 2021-09-09
GSP-749: Unify Path Behavior

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/744


docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,88 @@
...
+  - For Windows, the absolute path starts with a drive letter, `/` is the current drive.
+- Relative Path: 
+  - The relative path is for the working directory `WorkDir`.
+  - Path with prefix `./` or `../` is allowed.
+
+### WorkDir and path
+
+`WorkDir` specifies the working directory of the process.
+
+- For file system, `WorkDir` SHOULD be a passed in absolute path or relative to the current directory of the process.
```
JinnyYi, line 39: Do we need to restrict the `WorkDir` to be an absolute path?  So that for fs in Windows, `WorkDir` should be `c:\a\b`?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,88 @@
...
+  - The default value is `/`, that means the working directory is the root path for Unix, and the current drive of the process for Windows.
+  - Services should set `WorkDir` to the path name after the evaluation of any symbolic links internal.
+- For object storage, `WorkDir` is the simulated directory or prefix of the object key. 
+  - The default value is `""`.
+  - `WorkDir` SHOULD be the unix style and SHOULD NOT with prefix `/` when the value is not empty.
```
JinnyYi, line 44: This description is not consistent with the current implementation. `WorkDir` is an optional pair for `s3` in `site`, but we set the default value to `/` in our implementation. I'm not sure whether the working directory should not be empty for all the services.

docs/rfcs/749-unify-path-behavior.md
Xuanwo: > Do we need to restrict the `WorkDir` to be an absolute path?

Yes, I think we need to do this.

docs/rfcs/749-unify-path-behavior.md
Xuanwo: I think we should use the same default for workdir for all services. Even for s3, we need to use `/` as the workdir. And we will convert `/abc/def` to `abc/def` in `getAbsPath` insides services.

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,88 @@
...
+`path` is the file path for file system, or an object key for object storage. Also, it could be a prefix filter for `List` operation.
+
+- `path` could be an absolute path or a relative path.
+- Services SHOULD convert it to the absolute path at the beginning of the operation.
+- For the unique key `Object.ID` in storage, it should be an absolute path, unless there's a returned unique identifier like in dropbox.
+- Users SHOULD follow the file and object naming of different services.
+
+### Path Separator
+
+All the passed in path SHOULD be unix style, no matter Linux platform or Windows platform, no matter object storage service or file system.
```
Xuanwo, line 55: It is easy for services to implement, but hard to use. Users run on windows have to convert the path into unix path instead.

- Can we allow user to input path that contains windows path separator?
- Maybe we can do the convert in definitions? It that OK?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,88 @@
...
+When the drive letter is included for Windows, it should be something like `c:/a/b`.
```
Xuanwo, line 56: Maybe we only need to allow this case inside `go-service-fs`?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,88 @@
...
+
+From service side:
+
+- Services SHOULD convert `/` in path to the system-related path separator at the beginning of the operation.
```
Xuanwo, line 60: It could raise the problem is an app built upon go-storage running across different platforms.

- How about return unix style always?

docs/rfcs/749-unify-path-behavior.md
Xuanwo: I found it hard to handle, maybe we can limit `go-service-fs` only support current volume? So that we don't need to support `read("c:\\x\\y\\z")`?

docs/rfcs/749-unify-path-behavior.md
JinnyYi: Got it.

docs/rfcs/749-unify-path-behavior.md
JinnyYi: > Can we allow user to input path that contains windows path separator?

Yes, got it. And maybe services can also do the convert in case the user passes in the wrong format?

> Maybe we can do the convert in definitions? It that OK?

I don't quite understand the intention.

docs/rfcs/749-unify-path-behavior.md
JinnyYi: > I found it hard to handle, maybe we can limit `go-service-fs` only support current volume? So that we don't need to support read("c:\\x\\y\\z")?

Does it mean we don't need to support drive letter for all the paths including `WorkDir`? Or only for `WorkDir`(to point the working directory as the current volume like `c:\`, but we can write to `d:\test`)?

docs/rfcs/749-unify-path-behavior.md
JinnyYi: > How about return unix style always?

Perhaps this contradicts the Windows style path we are allowed？

Xuanwo:
Let's start a quick discussion in room: https://matrix.to/#/#beyondstorage@gsp-749:matrix.org

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Proposal
+
+### Absolute Path and Relative Path
+
+All our services should support two kinds of path:
+
+- Absolute Path:
+  - For Unix, the absolute path starts with `/`.
+  - For Windows, the absolute path starts with a drive letter.
```
Xuanwo, line 30: Maybe only `go-service-fs` will support absolute path starts with a drive letter?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+- Relative Path: 
+  - The relative path is for the working directory `WorkDir`.
+  - Path with prefix `./` or `../` is allowed: they are relative paths for file system, but common strings for object storage service.
```
Xuanwo, line 33: It may be hard to support. How about drop those support?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+  - The default value is `/`, that means the working directory is the root path for Unix, and the current volume of the process for Windows.
+  - Services SHOULD set `WorkDir` to the path name after the evaluation of any symbolic links internal.
+- For object storage, `WorkDir` is the simulated directory or prefix of the object key.
+  - `WorkDir` SHOULD be Unix style with the prefix `/`.
+  - The default value is `/`.
+  - Form service side, the prefix `/` needs to be removed for internal processing, but `/` needs to be added to the fields returned to the user.
+  
+`path` is the file path for file system, or an object key for object storage. Also, it could be a prefix filter for `List` operation.
+
+- For file system services on Windows, `path` MUST be the relative path for `WorkDir`. For file system on Unix, `path` could be an absolute path or a relative path.
```
Xuanwo, line 50: For `go-service-fs` on windows, absolute path is OK which means a path under current volume. But absolute path with driver letter is not allowed.

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+All our services should support two kinds of path:
+
+- Absolute Path:
+  - For Unix, the absolute path starts with `/`.
+  - For Windows, the absolute path starts with a drive letter.
+- Relative Path: 
+  - The relative path is for the working directory `WorkDir`.
+  - Path with prefix `./` or `../` is allowed: they are relative paths for file system, but common strings for object storage service.
+
+### WorkDir and path
```
Xuanwo, line 35: I found the current rule is too complex. Maybe we can present the general rules first and then leave exception rules to `go-service-fs` RFC?

docs/rfcs/749-unify-path-behavior.md
JinnyYi: Maybe we don't need to specify this explicitly, but leave it to the services to support it according to their own default rules?

docs/rfcs/749-unify-path-behavior.md
JinnyYi: Got it.

docs/rfcs/749-unify-path-behavior.md
JinnyYi: I‘ll reorganize the rules as your suggestion.

docs/rfcs/749-unify-path-behavior.md
Xuanwo: I thought we will not support `Read("C:\\xxxx")` ?

docs/rfcs/749-unify-path-behavior.md
Xuanwo: Maybe we can remove this sentence simply?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Absolute Path and Relative Path
+
+All our services should support two kinds of path:
+
+- Absolute Path:
+  - For Unix, the absolute path starts with `/`.
+  - For Windows, the absolute path starts with a drive letter.
+- Relative Path: 
+  - The relative path is for the working directory `WorkDir`.
```
Xuanwo, line 32: A bit strange. What's the meaning of `for`? How about make it more clear?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,89 @@
...
+  - The name cannot start with a forward slash (`/`) or a backslash (`\`).
+
+## Compatibility
+
+- `work_dir` must start with `/` is only for Unix.
+- The format for `name` and `work_dir` in connection string need to be reconsidered as we can't separate them by `/` anymore.
+
+## Implementation
+
+N/A
```
Xuanwo, line 89: Please give more detail of the implementation?

- Will we change the definitions?
- Will we add new integration tests?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,151 @@
...
+All our services should support two kinds of path:
+
+- Absolute path: must include the root directory.
+- Relative path: a relative path based on the working directory.
+
+### Directory separator
+
+From user side:
+
+- System-related directory separator including `\` for Unix and `//` for Windows SHOULD be allowed.
```
Xuanwo, line 34: Should be `/` for unix and `\\` for windows?

docs/rfcs/749-unify-path-behavior.md
```diff
@@ -0,0 +1,151 @@
...
+	store.workDir = opt.WorkDir
+}
+```
+
+**path**
+
+The field `path` is used as an input parameter for operation, indicating the file path for file system, object key for object storage or prefix filter for `List` operator.
+
+- `path` could be absolut path or relative path based on `work_dir`.
+- `path` MUST be Unix style path but SHOULD NOT contain the prefix `/` for object storage services.
```
Xuanwo, line 75: When users call `store.Read`, they don't know whether this storage is object storage or not. So I think it's not a good rule.

And it conflicts with the example in the following.

Xuanwo:
Mostly LGTM, let's create a tracking issue for this GSP!

PR #751, opened 2021-09-10
GSP-751: Write Empty File Behavior

abyss-w:
Write Empty File Behavior


docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,51 @@
+- Author:  Abyss-w <mad.hatter@foxmail.com>
+- Start Date: 2021-09-10
+- RFC PR: [beyondstorage/go-storage#751](https://github.com/beyondstorage/go-storage/pull/751)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
+
+# GSP-751: Write Empty File Behavior
+
```
JinnyYi, line 7: How about adding [Specify the behavior for writing empty file](https://forum.beyondstorage.io/t/topic/204) as the previous Discussion?

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,51 @@
...
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
+
+# GSP-751: Write Empty File Behavior
+
+## Background
+
+In our definition, the `Write` API will upload a file to the target object. We do not do anything about uploading empty files. However, some services do not support uploading empty files. As [proposed](https://forum.beyondstorage.io/t/topic/204) in our community, for s3, when we upload an empty file, the service will hang permanently. Here is an example of s3:
+
+```go
+_, err = fs.s.Write(path,nil, 0)
```
JinnyYi, line 13: Maybe `fs` should be `s3` to keep consistent with the description above?

abyss-w: Yes. It's better this way.

docs/rfcs/751-write-empty-file-behavior.md
abyss-w: I think it can be done like this.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,51 @@
...
+}
+```
+
+This is not convenient for the users.
+
+## Proposal
+
+I propose to allow the user to pass in an empty file when calling `Write`. 
+
+- For services that do not support uploading empty files, we should check if the `reader` is `nil`. If it is `nil`, we need to create a `reader` before calling the API. Like s3, kodo, etc.
```
JinnyYi, line 36: Maybe we should describe it differently instead of "empty file" as services like s3, kodo support upload or empty file or object with size 0, but `io.Reader` is not allowed to be `nil` for the `Write` APIs?

abyss-w: Got it!

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,59 @@
...
+}
+```
+
+This is not convenient for the users.
+
+## Proposal
+
+I propose to allow the user to pass in a nil `io.Reader` when calling `Write`.
+
+- For services that do not support uploading nil `io.Reader`, but support upload object with size 0 , we should check if `io.Reader` is `nil`. If it is `nil`, we need to create a `io.Reader` before calling the API. For example, `s3`, `kodo`, etc.
```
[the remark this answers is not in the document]
Xuanwo: In this proposal, I expect to see discussion about the following:

- What will happen if we got a nil `io.Reader` but `size != 0`?
- What will happen if we got a valid `io.Reader` but `size = 0`?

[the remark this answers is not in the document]
abyss-w: Got it!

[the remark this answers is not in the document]
abyss-w: I found that `azfile` does not support uploading a nil reader, nor does it support uploading a reader with size 0.
Then for example `kodo`, when the size and reader length are not the same, it will report an error. But most other services will automatically recognize the length of the reader when size is `0`.
Then most of the services will report an error when the length of the reader is `0` and the size is not `0`.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+
+# GSP-751: Write Empty File Behavior
+
+- Previous discussion:
+  - [Specify the behavior for writing empty file](https://forum.beyondstorage.io/t/topic/204)
+
+## Background
+
+```go
+func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error)
```
Xuanwo, line 14:
```suggestion
func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error) {}
```

Make github PR preview works as expected.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+In our definition, the `Write` API will upload a file to the path. We do not do anything about uploading empty files(i.e. `io.Reader`is nil ). However, some services do not support uploading nil `io.Reader`. As [proposed](https://forum.beyondstorage.io/t/topic/204) in our community, for s3, when we upload an empty file, the service will hang permanently. Here is an example of s3:
+
+```go
+_, err = s3.s.Write(path,nil, 0)
```
Xuanwo, line 20: Please format the code correctly.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+if err != nil {
+s3.logger.Error("write", zap.String("path", path), zap.Error(err))
+return nil, nil, err
+}
+```
+
+If we want to upload an empty file, we need to do so:
+
+```go
+_, err = s3.s.Write(path, bytes.NewReader([]byte{}), 0)
```
Xuanwo, line 30: ditto.

docs/rfcs/751-write-empty-file-behavior.md
Xuanwo: Maybe we can use `_, err = store.Write(path, nil, 0)` here so that our readers don't need to care about what `s3.s` is.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+```go
+func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error)
+```
+
+In our definition, the `Write` API will upload a file to the path. We do not do anything about uploading empty files(i.e. `io.Reader`is nil ). However, some services do not support uploading nil `io.Reader`. As [proposed](https://forum.beyondstorage.io/t/topic/204) in our community, for s3, when we upload an empty file, the service will hang permanently. Here is an example of s3:
+
+```go
+_, err = s3.s.Write(path,nil, 0)
+if err != nil {
+s3.logger.Error("write", zap.String("path", path), zap.Error(err))
```
Xuanwo, line 22: Maybe we can use `return err` here to make our demo more clear.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+return nil, nil, err
+}
+```
+
+This is not convenient for the users.
+
+## Proposal
+
+I propose to allow the user to pass in a nil `io.Reader` and `0` size to create an empty file when calling `Write`.
```
Xuanwo, line 41: Please make the behavior more clear in the proposal instead of in `Possible Q&As`.

> NOTES: `Rationale` is used to add more extra information, but the behavior is the most important part of GSP, we can't separate them to `Rationale`.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Rationale
+
+### Possible Q&As
+
+- What will happen if we got a nil `io.Reader` but `size != 0`?
```
Xuanwo, line 47: The behavior is related to https://forum.beyondstorage.io/t/topic/190

NOTES: we can't make decisions for users.

If a user tries to upload a file with 10 bytes, but pass a nil `io.Reader`, we need to return an error instead change the size.

Can you re-think the behavior of this GSP?

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+  - If the upload is successful, we will return the size as the length of the `io.Reader`.
+
+## Compatibility
+
+This behavior will affect all services that do not support uploading empty files.
+
+## Implementation
+
+- Update definitions to reflect changes
+- Update integrations tests to make sure all service passed
```
Xuanwo, line 61: Please make it clear that what tests we will add.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+  - We will upload files of `io.Reader` length.
+  - If the upload is successful, we will return the size as the length of the `io.Reader`.
+
+## Compatibility
+
+This behavior will affect all services that do not support uploading empty files.
+
+## Implementation
+
+- Update definitions to reflect changes
```
Xuanwo, line 60: Do we need to update definitions?

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,62 @@
...
+- What will happen if we got a nil `io.Reader` but `size != 0`?
+  - We will set the size to `0` before calling the API.
+  - If the API call is successful we will return size `0`.
+- What will happen if we got a valid `io.Reader` but `size = 0`?
+  - We will upload files of `io.Reader` length.
+  - If the upload is successful, we will return the size as the length of the `io.Reader`.
+
+## Compatibility
+
+This behavior will affect all services that do not support uploading empty files.
```
Xuanwo, line 56: Please make it clear that it's a compatible change.

abyss-w:
The above has been received and I will make the changes.

docs/rfcs/751-write-empty-file-behavior.md
abyss-w: Maybe we should add this behavior inside the description of `Write`?

Xuanwo: Makes sense, please make it more clear.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,77 @@
...
+## Rationale
+
+N/A
+
+## Compatibility
+
+This change will not break services and users. We can do this as follows:
+
+- Add this behavior to `Write` in `go-storage`.
+- Services bump to the latest version of `go-storage' (if the latest version does not contain this behavior, bump to the latest master branch), and all descriptions of this behavior are updated. Then changes are made to the services that need to be modified.
```
Xuanwo, line 64: We don't need to require a version bump in GSP, we will bump services after the go-storage released a new version.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,77 @@
...
+```go
+func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error) {}
+```
+
+In our definition, the `Write` function will upload a file to the path. We do not do anything about uploading empty files(i.e. `io.Reader`is nil ). However, some services do not support uploading nil `io.Reader`. As [proposed](https://forum.beyondstorage.io/t/topic/204) in our community, for s3, when we upload an empty file, the service will hang permanently. Here is an example of s3:
+
+```go
+_, err = store.Write(path,nil, 0)
+if err != nil {
+return err
```
Xuanwo, line 22:
```suggestion
    return err
```

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Proposal
+
+I propose to allow the user to pass in a nil `io.Reader` and `0` size to create an empty file when calling `Write`. For services that do not support nil Reader uploads we can do something like this.
+
+```go
+func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error) {
+// We can create a Reader with size 0 like this.
+r = bytes.NewReader([]byte{})
+// Then we can call the API with r and 0 size.
```
Xuanwo, line 45:
```suggestion
    // We can create a Reader with size 0 like this.
    r = bytes.NewReader([]byte{})
    // Then we can call the API with r and 0 size.
```

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+- What will happen if we got a nil `io.Reader` but `size != 0`?
+  - We will return an error. We can't make decisions for users.
+- What will happen if we got a valid `io.Reader` but `size = 0`?
+  - We will upload files of `io.Reader` length.
```
Xuanwo, line 52: `size` is the limit size of this write operation, and we will not try to get the `io.Reader`'s size.

So in this case, I think we will create an empty file too.

abyss-w: Got it!

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,85 @@
...
+
+```go
+func (s *Storage) Write(path string, r io.Reader, size int64, pairs ...Pair) (n int64, err error) {
+    // We can create a Reader with size 0 like this.
+    r = bytes.NewReader([]byte{})
+    // Then we can call the API with r and 0 size.
+}
+```
+
+- What will happen if we got a nil `io.Reader` and `size = 0`?
```
Xuanwo, line 49: How about writing them like a markdown table?

| reader | size |  description |
| - | - | - |
| valid | zero | xxxx |
| valid | valid | xxxx |
| nil | zero | xxxx |
| nil | valid | xxxx |

The markdown syntax is:


```markdown
| reader | size |  description |
| - | - | - |
| valid | zero | xxxx |
| valid | valid | xxxx |
| nil | zero | xxxx |
| nil | valid | xxxx |
```

If the table is not good for long sentences, how about:

## valid reader and valid size

xxx

## valid reader and zero size

xxx

## nil reader and valid size

xxx

## nil reader and zero size

xxx

And at last, we can give a conclusion about all those cases.

abyss-w: Got it!

abyss-w: I think the second way is better.

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,94 @@
+- Author:  Abyss-w <mad.hatter@foxmail.com>
+- Start Date: 2021-09-10
+- RFC PR: [beyondstorage/go-storage#751](https://github.com/beyondstorage/go-storage/pull/751)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
```
Xuanwo, line 4: Mostly LGTM, let's create a tracking issue and upload the link here!

docs/rfcs/751-write-empty-file-behavior.md
```diff
@@ -0,0 +1,94 @@
...
+Next are some situations that users will encounter when calling `Write`.
+
+### nil io.Reader and zero size
+
+- As described above, we will create a `io.Reader` of size 0 and upload it.
+- We will upload data of length `size` as long as the length of `io.Reader` is the same as `size`.
+
+### nil io.Reader and valid size
+
+- We will return an error. In this case, the user's action is wrong, so we should return an error to alert the user.
```
JinnyYi, line 58: Maybe we need to add a new error code for it?

Xuanwo: I think it's the same error that should be covered in https://forum.beyondstorage.io/t/topic/190, how about return error directly and formalize along with next proposal?

Issue #760, opened 2021-09-18
Compilation fails if `Default*Pairs` unimplemented in service


JinnyYi:
### What happened?

A compilation error occured when service does not implement `Default*Pairs`:

```shell
vet: ./generated.go:145:10: result.HasDefaultStoragePairs undefined (type pairStorageNew has no field or method HasDefaultStoragePairs)
```

ref: https://github.com/beyondstorage/go-service-example/runs/3497258311?check_suite_focus=true

### Version

4.7.0

Xuanwo:
Let's fix this problem in this sprint: https://github.com/beyondstorage/community/issues/48

PR #821, opened 2021-09-22
fix: Compilation fails if Default*Pairs unimplemented in service

JinnyYi:
fix https://github.com/beyondstorage/go-storage/issues/760


Xuanwo:
Maybe `default_storage_pairs` and `storage_features` should be required?

How about always add `default_storage_pairs` and `storage_features` for services?

JinnyYi:
> Maybe `default_storage_pairs` and `storage_features` should be required?
> 
> How about always add `default_storage_pairs` and `storage_features` for services?

Does it mean `default_storage_pairs` should be added in `namespace.storage.new`  for all services? Or should we generate `DefaultStoragePairs` in `pairStorageNew`  struct no matter there's defalt pair or not?

For `storage_features`, no additional logic is currently required to handle it.

Xuanwo:
> Does it mean `default_storage_pairs` should be added in `namespace.storage.new` for all services?

Yes, I think this way could be better.

JinnyYi:
> > Does it mean `default_storage_pairs` should be added in `namespace.storage.new` for all services?
> 
> Yes, I think this way could be better.

Then maybe we don't need to modify `go-storage` for `default_storage_pairs`.

Xuanwo:
> Then maybe we don't need to modify `go-storage` for `default_storage_pairs`.

How about adding them by default? And finally, we can remove `default_storage_pairs` in `services.toml` as they will be auto-added.

tests/generated.go
```diff
@@ -163,14 +163,8 @@ type pairServiceNew struct {
 	HasCredential bool
 	Credential    string
 	// Optional pairs
-	HasDefaultContentType  bool
```
Xuanwo, line 166: Why they are been removed?

JinnyYi: There's no operation in `service` namespace has the `content_type` pair. So I modified the way to get the defaultable pairs in https://github.com/beyondstorage/go-storage/pull/821/files#diff-0ca1d3b96d3cc7d55c46d876668a196e55b5a73e0208c01307c5d93e93d0bc9cR311-R312

Xuanwo: Makes sense.

cmd/definitions/type.go
```diff
@@ -307,14 +308,34 @@ func (d *Data) LoadService(filePath string) {
...
+			ns.New.Optional = append(ns.New.Optional, "default_"+v.Pair.Name)
+		}
+		// add default_*_pairs and *_features
+		hasDefaultNsPairs := false
+		hasNsFeatures := false
+		defaultNsPairsName := fmt.Sprintf("default_%s_pairs", ns.Name)
+		nsFeaturesName := fmt.Sprintf("%s_features", ns.Name)
+		for _, v := range ns.New.Optional {
+			if v == defaultNsPairsName {
+				hasDefaultNsPairs = true
```
Xuanwo, line 321: How about adding a warning here to let maintain remove the `default_storage_pairs` and `storage_features`? As we will auto generate for them.

Xuanwo (approved):

cmd/definitions/type.go
JinnyYi: Maybe I also should remove the pairs from tests.

PR #824, opened 2021-09-26
fix: Not enough arguments in call to Service.formatError

JinnyYi:
## Issue
Compilation error occured when bump go-storage in go-service-*:
```shell
# github.com/beyondstorage/go-service-qingstor/v3
./generated.go:426:17: not enough arguments in call to s.formatError
        have (string, error)
        want (string, error, string)
# github.com/beyondstorage/go-service-qingstor/v3
vet: ./generated.go:426:29: too few arguments in call to s.formatError
```
## cause
Not generate enough parameters in call to `formatError` for `List` in `Service`.
ref: https://github.com/beyondstorage/go-storage/blob/master/tests/generated.go#L409-L412

### DIFF
--- a/cmd/definitions/gen_service.go
+++ b/cmd/definitions/gen_service.go
@@ -486,12 +486,12 @@ If user enable this feature, service should ignore not support pair error.`),
 				gg.Defer(gg.Embed(func() gg.Node {
 					caller := gg.Call("formatError").WithOwner("s")
 					caller.AddParameter(gg.Lit(fn.Name)).AddParameter("err")
-					for _, v := range op.ParsedParams() {
-						// formatError only accept string as input.
-						if v.Type != "string" {
-							continue
-						}
-						caller.AddParameter(v.Name)
+					paths := op.PathCaller()
+					if len(paths) == 0 && nsNameP == "Service" {
+						caller.AddParameter("\"\"")
+					}
+					for _, v := range paths {
+						caller.AddParameter(v)
 					}
 
 					fn := gg.Function("").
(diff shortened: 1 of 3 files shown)


cmd/definitions/type.go
```diff
@@ -653,6 +654,19 @@ func (op *Operation) ParsedResults() []*Field {
 	return fs
 }
 
+func (op *Operation) PathCaller() []string {
```
Xuanwo, line 657: It's so wired for us to still have a `PathCaller`....

cmd/definitions/gen_service.go
```diff
@@ -486,12 +486,12 @@ If user enable this feature, service should ignore not support pair error.`),
 				gg.Defer(gg.Embed(func() gg.Node {
 					caller := gg.Call("formatError").WithOwner("s")
 					caller.AddParameter(gg.Lit(fn.Name)).AddParameter("err")
-					for _, v := range op.ParsedParams() {
-						// formatError only accept string as input.
-						if v.Type != "string" {
-							continue
-						}
-						caller.AddParameter(v.Name)
+					paths := op.PathCaller()
```
Xuanwo, line 489: `formatError` is well defined, maybe we can add parameters directly?

cmd/definitions/gen_service.go
```diff
@@ -486,12 +486,18 @@ If user enable this feature, service should ignore not support pair error.`),
 				gg.Defer(gg.Embed(func() gg.Node {
 					caller := gg.Call("formatError").WithOwner("s")
 					caller.AddParameter(gg.Lit(fn.Name)).AddParameter("err")
+					paramsNum := 2
```
Xuanwo, line 489: It looks like we only need a `isEmpty := true` here, do we really need to count the nums?

JinnyYi: No, the num is unnecessary.

Xuanwo (approved):

Xuanwo:
Looks good now.

PR #825, opened 2021-09-26
fix: Replace each backslash in path with a slash

JinnyYi:


cmd/definitions/gen_service.go
```diff
@@ -529,7 +529,7 @@ If user enable this feature, service should ignore not support pair error.`),
 								continue
 							}
 							if v.Name == "path" || v.Name == "src" || v.Name == "dst" || v.Name == "target" {
-								ic.AddParameter(gg.S("filepath.ToSlash(%s)", v.Name))
+								ic.AddParameter(gg.S("strings.ReplaceAll(%s, \"\\\\\", \"/\")", v.Name))
```
Xuanwo, line 532: Maybe we can use 

```go
gg.S(`strings.ReplaceAll(%s, "\\", "/")`, v.Name)
```

tom improve the readability?

PR #826, opened 2021-09-26
GSP-826: Add Multipart HTTP Signer Support

abyss-w:


abyss-w:
I have opened a draft PR, please point out any problems, and I will fix them.

Xuanwo:
We need to support `Delete` in signer too: https://forum.beyondstorage.io/t/topic/226

abyss-w:
> We need to support `Delete` in storage_http_signer too: https://forum.beyondstorage.io/t/topic/226

Got it!

docs/rfcs/826-add-multipart-http-signer-support.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Proposal
+
+I propose to add `MultipartHTTPSigner` interface.
+
+```go
+type MultipartHTTPSigner interface {
+signHTTPCreateMultipart(path string, expire time.Duration, ps ...types.Paires) (req *http.Request, err error)
+signHTTPWriteMultipart(o *Object, size int64, index int, expire time.Duration, ps ...types.Pairs) (req *http.Request, err error)
+}
```
JinnyYi, line 27: The added interfaces should be public. How about `QuerySignHTTPCreateMultipart` and `QuerySignHTTPWriteMultipart`? And the format(indent) should be updated.

docs/rfcs/826-add-multipart-http-signer-support.md
```diff
@@ -0,0 +1,70 @@
...
+
+- There's no appropriate way to pass in some parameters for some operations like [How to pass partIndex into QuerySignHTTP for WriteMultipart](https://forum.beyondstorage.io/t/how-to-pass-partindex-into-querysignhttp-for-writemultipart/192).
+- `CreateMultipart` needs to support Expires parameter.
+
+## Proposal
+
+I propose to add `MultipartHTTPSigner` interface.
+
+```go
+type MultipartHTTPSigner interface {
```
JinnyYi, line 24: We should also support `ListMultipartUploads` and `CompleteMultipart`.

docs/rfcs/826-add-multipart-http-signer-support.md
```diff
@@ -0,0 +1,78 @@
...
+    QuerySignHTTPCreateMultipart(path string, expire time.Duration, ps ...types.Pairs) (req *http.Request, err error)
+    QuerySignHTTPWriteMultipart(o *Object, size int64, index int, expire time.Duration, ps ...types.Pairs) (req *http.Request, err error)
+    QuerySignHTTPListMultipart(o *Object, expire time.Duration, ps ...types.Pairs) (req *http.Request, err error)
+    QuerySignHTTPCompleteMultipart(o *Object, parts []*Part, expire time.Duration, ps ...types.Pairs) (req *http.Request, err error)
```
JinnyYi, line 28: Where is `o *Object` come from?

abyss-w: Oh, but we need the `path` and `uploadId`. How can we pass them in better?

JinnyYi: I think we need to add return value for `QuerySignHTTPCreateMultipart`, maybe `MultipartID` or `Object`? And also need to add return value for `QuerySignHTTPWriteMultipart`(including `Part`).

Xuanwo: I think the current design is OK. Users can use `Create` to build an object.

JinnyYi: `MultipartID` and `Part`(`ETag`) are returned by server. If the current design is kept, we will not be able to use `QuerySignHTTPCreateMultipart` and `QuerySignHTTPWriteMultipart`  at the same time, and the part uploaded by `QuerySignHTTPWriteMultipart` will not be able to merge due to the lack of etag.

JinnyYi: Oh, I misunderstood it. We just constructed the `*http.Request` for users and the requests are not send so we can't get these infos.

Xuanwo: Yes, they can't. They need to parse the returning response by themself.

For example:

```go
req, _ := store.QuerySignHTTPCreateMultipart(xxxx)
resp, _ := client.Send(req)
var cmr createMultipartResp
_ := json.Unmarshal(resp.Body, &cmr)

o := store.Create(xxxx, cmr.MustGetMultipartId())
req, _ = store.QuerySignHTTPWriteMultipart(o, xxx)
```

Xuanwo:
Mostly LGTM, please create a tracking issue before @JinnyYi makes the final decision.

JinnyYi:
I have one more question. Do we need to add `QuerySignHTTPAbortMultipart` instead of supporting it in `QuerySignHTTPDelete`？ If we keep `QuerySignHTTPDelete`, should `List`(`QuerySignHTTPList` instead of `QuerySignHTTPListMultipart`) also be consistent?

Xuanwo:
> Do we need to add `QuerySignHTTPAbortMultipart` instead of supporting it in `QuerySignHTTPDelete`？

No, I think we don't need to. As `Delete` with multipart id is the same with `AbortMultipart`.

> If we keep `QuerySignHTTPDelete`, should `List`(`QuerySignHTTPList` instead of `QuerySignHTTPListMultipart`) also be consistent?

I think they are different. `List` supports `ListModePart` to list Multipart objects, but `ListMultipart` is used to list parts inside a multipart object.

docs/rfcs/826-add-multipart-http-signer-support.md
```diff
@@ -0,0 +1,78 @@
+- Author: abyss-w <mad.hatter@foxmail.com>
+- Start Date: 2021-09-26
+- RFC PR: [beyondstorage/go-storage#826](https://github.com/beyondstorage/go-storage/pull/826)
+- Tracking Issue: [beyondstorage/go-storage#0](https://github.com/beyondstorage/go-storage/issues/0)
```
Xuanwo, line 4: Please update the tracking issue link here

abyss-w: Done

JinnyYi (approved):

PR #847, opened 2021-10-14
refactor(tests): Move go-integration-tests here

Xuanwo:


Xuanwo:
After this change, we can implement features and tests together!

PR #879, opened 2021-10-18
feat(services/bos): Move services bos back

Joey-1445601153:
Move services bos back to go-storage


Xuanwo:
Please also update https://github.com/beyondstorage/go-storage/blob/master/.github/dependabot.yml

PR #884, opened 2021-10-18
feat(services/cos): Move services cos back

Joey-1445601153:
Move services cos back to go-storage


Xuanwo:
Please update your PR's title.

PR #895, opened 2021-10-19
feat(services/ftp): Move services ftp back

Joey-1445601153:
Move services ftp back to go-storage/services


Joey-1445601153:
@Xuanwo help.  
Afer I changed the integration-test config, now it can work. But there are lots of broken pipe errors.
And I don't know what's wrong.

Xuanwo:
> @Xuanwo Help! Afer I changed the integration-test config, now it can work. But there are lots of broken pipe errors. And I don't know what's going on.

Oh, it causes by our testing env.

Let me find a solution.

PR #901, opened 2021-10-19
docs(services/gcs): Add status badge for README

Joey-1445601153:
Add badge for README.md


Xuanwo:
There is not change in this PR, maybe something wrong?

Joey-1445601153:
> There is not change in this PR, maybe something wrong?
Just fancy! Normally if there's no change, how can I start a pr?Maybe i should close this and start another?

PR #916, opened 2021-10-20
ci(*): Upgrade minimum version to Go 1.16

Juneezee:
1. Drop Go 1.15 and add Go 1.17 to GitHub Actions workflows.
~2. Remove `tidy` as the dependency of `build` target and add another `make tidy-all` step to `build-test.yml`. This is required otherwise `make build-all` will report an error when the current go.mod file indicates a higher Go version than the current installed version.~
3. The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). This PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

EDIT: Revert Makefile changes

### DIFF
--- a/cmd/definitions/parse.go
+++ b/cmd/definitions/parse.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	"github.com/Xuanwo/templateutils"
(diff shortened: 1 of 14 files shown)


Xuanwo:
Thanks for your contributions as first.

We have set a GSP about our release policy: https://github.com/beyondstorage/go-storage/blob/master/docs/rfcs/15-release-policy.md

So upgrade to 1.17 is not suitable for us. 

Upgrade to 1.16 is OK though.

We have a discussion about it at our forum:https://forum.beyondstorage.io/t/topic/244

Do you want to implement it?

Xuanwo:
BTW, it's OK to me to just add go 1.17 support in this PR. And we can discuss the upgrade to 1.16 in another one.

Xuanwo:
Yes, we can drop go 1.15 in github action. 😝

Xuanwo:
Maybe we don't need to change the Makefile if we use `go 1.16` in go.mod?

PR #917, opened 2021-10-20
feat(services/obs): Move services obs back

Joey-1445601153:
Move services obs back to go-storage


JinnyYi:
> We don't have a valid obs test account now. Let's add an issue to track and make this PR merge.

Got it. Maybe we can remove the obs test workflow in this PR and add it when the test account is ready?

Issue #920, opened 2021-10-20
Tracking issue of services tests


Xuanwo:
## No Test Account

- [ ] https://github.com/beyondstorage/go-storage/issues/918
- [ ] #923

## Not Stable

- [ ] https://github.com/beyondstorage/go-storage/issues/906

Joey-1445601153:
## No Test Account
> * [ ]  [services/onedrive: Integration tests failed because of lacking onedrive test account. #923](https://github.com/beyondstorage/go-storage/issues/923)

Xuanwo:
Use https://github.com/beyondstorage/go-storage/issues/536 to track instead.

Issue #918, opened 2021-10-20
services/obs: Integration tests failed because of lacking obs test account.


Joey-1445601153:
`
Run 1password/load-secrets-action@v1
Run export INPUT_UNSET_PREVIOUS=false
Populating variable: STORAGE_OBS_ENDPOINT
Getting vault ID from vault name: Engineering
Getting item ID from vault cdmcu75h3o7xdjczun7fbbsmui...
Loading item null from vault cdmcu75h3o7xdjczun7fbbsmui...
curl: (22) The requested URL returned error: 400 Bad Request
Error: Process completed with exit code 22.
`

PR #922, opened 2021-10-20
feat(services/onedrive): Move services onedrive back

Joey-1445601153:
Move services onedrive back to go-storage.

### DIFF
--- a/docs/rfcs/10-callback-reader.md
+++ b/docs/rfcs/10-callback-reader.md
@@ -27,7 +27,7 @@ After all these work, we can work well with progress bar:
 
 ```go
 import (
-	"io"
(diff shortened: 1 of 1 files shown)


Joey-1445601153:
@Xuanwo  I check the 1password, we don't have onedrive test account too.So it's a problem that has happend.
Maybe we should handle it like obs?

Xuanwo:
> @Xuanwo I check the 1password. we don't have onedrive test account too.So it's a problem that has happend? Maybe we should handle it like obs?

Yes. Can you create an issue about it like #918 and leave a comment at https://github.com/beyondstorage/go-storage/issues/920?

Issue #860, opened 2021-09-29
services/s3: Integration test fails when using the latest MinIO


JinnyYi:
## Excepted Behavior

`go-service-s3` should compatible with `MinIO` and pass integration test with `MinIO`.

## Current Behavior

- Integration test passed when using `minio RELEASE.2021-06-17T00-10-46Z`
- Decode REST XML response failed when using `MinIO Console`
  ```shell
  When Read a file     storager.go:55: write: Storager s3 {Name: xxxxx, WorkDir: /dc2c38b3-2de2-413c-8399-92cb1ba879e8/}, [3dd549ed-b614-45b7-9848-d33fdac98238]: unexpected: SerializationError: failed to decode REST XML response
                status code: 200, request id:
        caused by: XML syntax error on line 1: element <link> closed by </head>
  ```

## Steps to Reproduce

Run integration test for go-service-s3 with the following config:

```
STORAGE_S3_INTEGRATION_TEST=on
STORAGE_S3_CREDENTIAL=hmac:minioadmin:minioadmin
STORAGE_S3_ENDPOINT=https:play.min.io:9443
STORAGE_S3_NAME=<bucketname>
STORAGE_S3_LOCATION=local
```

Also we need to set `endpoint` for initialization:
```go
pairs.WithEndpoint("https:play.min.io:9443")
````

## Environment

- MinIO Console

PR #937, opened 2021-10-21
fix(services/s3): Support absolute path and idempotent delete

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/860


Xuanwo:
Oh, minio has documented this behavior:

https://github.com/minio/minio/blob/master/cmd/fs-v1-multipart.go#L801-L804

```go
// Implements S3 compatible Abort multipart API, slight difference is
// that this is an atomic idempotent operation. Subsequent calls have
// no affect and further requests to the same uploadID would not be
// honored.
```

Maybe we can refer them too.

Issue #966, opened 2021-10-26
Tracking issue of GSP-837: Support Feature Flag


JinnyYi:
- [Rendered GSP](https://github.com/beyondstorage/go-storage/pull/837)

## Steps

- [x] Add operations back to `Storager`
  - Test should not use type assertions.
- [x] Support feature flag API
  - Add `write_empty_object` in `features.toml` and add `virtual` property for virtual features.
  - Generate `ServiceFeatures` and `StorageFeatures` on go-storage side. Rename the current generated `ServiceFeatures` and `StorageFeatures` pairs and structs for services.
- [ ] Integration test supports feature flag
  - [tests: Migrate all tests to Feature and suite based](https://github.com/beyondstorage/go-storage/issues/1022)
- [x] Services migrate to go-storage
  - Update dot-import for the `types` package

PR #980, opened 2021-10-29
feat(tests): Run test cases based on features

JinnyYi:
ref: https://github.com/beyondstorage/go-storage/issues/966


Xuanwo:
Do you think it's a big burden to remove all deprecated test cases from services? Can we just remove them?

JinnyYi:
> Do you think it's a big burden to remove all deprecated test cases from services? Can we just remove them?

If we remove them directly, we need to remove the test functions called in services before releasing a new version.  Maybe we can do it together in https://github.com/beyondstorage/go-storage/pull/986?

Xuanwo:
Got it. Let's revisit this PR until we figure out what's the best way to organize those tests via features.

Issue #977, opened 2021-10-28
Add a new meta to StorageMeta to hold the service name


abyss-w:
In BeyondCTL, we need to get the name of the service when implementing stat, but there is no such metain `StorageMeta` to store the service name.
We can add a meta to `StorageMeta` that holds the name of the service.

JinnyYi:
`service` field has been added to `StorageMeta`, let's assign the value after or during the service migration.

PR #992, opened 2021-11-04
Add field in StorageMeta to hold the service name

Joey-1445601153:
Add field in StorageMeta to hold the service name

Fix wrong function usage in init function.

[original issue](https://github.com/beyondstorage/go-storage/issues/977)

### DIFF
--- a/definitions/infos.go
+++ b/definitions/infos.go
@@ -196,7 +196,7 @@ var InfosStorageMetaArray = []Info{
 		Name:        "service",
 		Type:        Type{Name: "string"},
 		export:      true,
-		Description: "Name of Services",
+		Description: "Service name.",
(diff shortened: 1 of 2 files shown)


definitions/infos.go
```diff
@@ -191,6 +192,12 @@ var InfosStorageMetaArray = []Info{
 		Type:        Type{Name: "int64"},
 		Description: "Maximum size for write operation.",
 	},
+	{
+		Name:        "service_name",
```
Xuanwo, line 196: How about use `service` here?

Joey-1445601153: Yes,a string named service is enough.

JinnyYi (approved):

Issue #959, opened 2021-10-23
services/obs: Service test failed for content mismatch


Xuanwo:
- Related PR: https://github.com/beyondstorage/go-storage/pull/957
- Test Case: https://github.com/beyondstorage/go-storage/runs/3982686814?check_suite_focus=true

```go
     When Read a file with offset or size 
      When Read with offset 
        The error should be nil ✔✔
        The content should be match ✔✘✔
      When Read with size 
        The error should be nil ✔✔
        The content should be match ✔✘✔
      When Read with offset and size 
        The error should be nil ✔✔
        The content should be match ✔✘✔

 Failures:

  * /data/actions/beta/go-storage/go-storage/services/obs/tests/storage_test.go 
  Line 14:
  Expected: '138112'
  Actual:   '451281'
  (Should be equal)

  * /data/actions/beta/go-storage/go-storage/services/obs/tests/storage_test.go 
  Line 14:
  Expected: '1288953'
  Actual:   '2688928'
  (Should be equal)

  * /data/actions/beta/go-storage/go-storage/services/obs/tests/storage_test.go 
  Line 14:
  Expected: '49793'
  Actual:   '291284'
  (Should be equal)
```

Obs implementation must be buggy, let's fix it before release.

Xuanwo:
Wait for https://github.com/huaweicloud/huaweicloud-sdk-go-obs/issues/10

PR #960, opened 2021-10-24
fix(services/obs): Service test failed for content mismatch

Xuanwo:
Close #959 

Signed-off-by: Xuanwo <github@xuanwo.io>

PR #996, opened 2021-11-08
services/obs: Support offset and size for read operation

Joey-1445601153:
[original issue 959](https://github.com/beyondstorage/go-storage/issues/959)


Xuanwo:
Hi, we have fixed this problem in PR https://github.com/beyondstorage/go-storage/pull/960

PR #997, opened 2021-11-08
feat(services/s3): Migrate to Factory init

Xuanwo:


Xuanwo:
ping @JinnyYi & @abyss-w for review.

services/s3/generated.go
```diff
@@ -82,269 +73,288 @@ func GetStorageSystemMetadata(s *StorageMeta) StorageSystemMetadata {
...
+	s.Stat = true
+	s.Write = true
+	s.WriteMultipart = true
+	if f.EnableVirtualDir {
+		s.VirtualDir = true
 	}
-	return result, nil
+	if f.EnableVirtualLink {
+		s.VirtualLink = true
+	}
```
JinnyYi, line 335: Actually `write_empty_obejct` is also allowed. But I think maybe it should not be initialized with `WithEnableWriteEmptyObject` as it's the service capability?

Xuanwo: Nice catch, I will find a way to distinguish them.

services/s3/utils.go
```diff
@@ -69,46 +68,58 @@ func (s *Storage) String() string {
 
 // New will create both Servicer and Storager.
 func New(pairs ...typ.Pair) (typ.Servicer, typ.Storager, error) {
-	return newServicerAndStorager(pairs...)
+	f := Factory{}
```
JinnyYi, line 71: Do we need to call `RegisterFactory` at a right place? As I got `service not registered` when using `NewStoragerFromString`.

Xuanwo: Yep, the `init()` is missing, I will add it back.

services/s3/generated.go
Xuanwo: Fixed in https://github.com/beyondstorage/go-storage/pull/1008

services/s3/utils.go
Xuanwo: Fixed in https://github.com/beyondstorage/go-storage/pull/1009

Xuanwo:
ping @JinnyYi to review again.

services/s3/generated.go
```diff
@@ -82,269 +73,289 @@ func GetStorageSystemMetadata(s *StorageMeta) StorageSystemMetadata {
...
+			xs := strings.SplitN(partService, "@", 2)
+			f.Credential, f.Endpoint = xs[0], xs[1]
+		}
+	}
+	if partStorage != "" {
+		slash := strings.IndexByte(partStorage[1:], '/')
+		if slash == -1 {
+			f.Name = partStorage[1:]
+		} else {
+			f.Name, f.WorkDir = partStorage[1:slash], partStorage[slash:]
```
JinnyYi, line 227: I'm not sure the connection string format. `s3://hmac:ak:sk@http:xxx:xxx/bucket/dir/?enable_virtual_dir` and `s3://bucket/dir/?endpoint=http:xxx:xxx&credential=hmac:ak:sk&enable_virtual_dir` both failed for `slice bounds out of range [1:0]` here.

Xuanwo: Oh, the input connection should trim the `s3://` prefix. I will add unit test and docs for this in next PR.

Xuanwo: Fixed in https://github.com/beyondstorage/go-storage/pull/1014

JinnyYi (approved):

Issue #1000, opened 2021-11-08
tests: List with ListModePart should be on specific path


Xuanwo:
https://github.com/beyondstorage/go-storage/blob/58e73a0d974c9b95a6ff8ff3d47b1f25c4cfb033/tests/multiparter.go#L246-L249

We need to use `store.List(path, pairs.WithListMode(types.ListModePart))` here.

Joey-1445601153:
Fixed by https://github.com/beyondstorage/go-storage/pull/1010.

Xuanwo:
@Joey2132, instead of add comment `Fixed by #1010.` here, you can say `Fixed #1000` in #1000.

Issue #924, opened 2021-10-21
services/s3: `List` with `types.ListModePart` mode test fails when using MinIO


JinnyYi:
`services/s3` should compatible with `MinIO` and pass integration test with `MinIO`:  `List` with `types.ListModePart` mode should lists in-progress multipart uploads.

1. `CreateMultipart` with key `420d60bd-c35a-4106-a347-4eaeaef5276e/6ed57725-4f94-4cf2-b2eb-81a857a1a3e3`: get the `MultipartID`: `dbacfbb1-5c55-43a9-aeac-ba0c39521c9e`
2. `WriteMultipart` with the `MultipartID` get from step 1 to write a part
3. `List` with `types.ListModePart` mode to get the iterator `it`.

## Excepted Behavior

We can retrieve all the in-progress multipart uploads by calling `Next()` continuously until get the `IterateDone`.
```
mo, err := it.Next()
mid, ok := mo.GetMultipartID()
// mid should equal to the `MultipartID` returned by `CreateMultipart`
```

## Current Behavior

When calling `it.Next()`:
```
runtime error: invalid memory address or nil pointer dereference
```

The `ListMultipartUploads` request and response are as follows, no in-progress mulitpart upload info in the response body:

![image](https://user-images.githubusercontent.com/82445294/138200191-b74066e4-ebc6-420a-b170-1bf1c64d8d0b.png)

## Steps to Reproduce

Run integration test for services/s3 with the following config:

```
STORAGE_S3_INTEGRATION_TEST=on
STORAGE_S3_CREDENTIAL=hmac:admin:password
STORAGE_S3_NAME=<bucketname>
STORAGE_S3_LOCATION=local
```

Also we need to set `endpoint` and `UsePathStyle` for initialization:
```go
pairs.WithEndpoint("http:<MinIO server IP>:9000")
s3.WithForcePathStyle()
````

## Environment

- MinIO: minio version RELEASE.2021-10-13T00-23-17Z
- services/s3: v2.5.0

JinnyYi:
Possible answer: https://github.com/minio/minio/issues/11686

Xuanwo:
I think the problem should be https://github.com/beyondstorage/go-storage/issues/1000

PR #1001, opened 2021-11-08
feat(services/s3): Remove service test on minio for s3

Joey-1445601153:
Currently service test on minio for s3 can't pass because of [issue 924](https://github.com/beyondstorage/go-storage/issues/924). We have to revoke it.
We will solve this problem in [issue 1000](tests: List with ListModePart should be on specific path #1000).


Xuanwo:
Hi, it's OK to just paste a GitHub link here. We don't need to use the style like `[issue 924](xxxx)`

Joey-1445601153:
> just

Got it.

PR #1013, opened 2021-11-10
feat(tests): ListMultipart on path instead of empty

Joey2132:
Give store.List a specific path. According to [minio/minio#11686](https://github.com/minio/minio/issues/11686), List multipart uploads only return values for exact object name.


Xuanwo:
It's hard to test whether it works or not, let's continue this work after #997 merged.

Joey2132:
> It's hard to test whether it works or not, let's continue this work after #997 merged.

Got it.

Xuanwo:
Please merge master and test again.

PR #1015, opened 2021-11-10
services/s3: Add tests upon minio

Joey2132:
Add tests upon minio for s3.


Xuanwo:
service test s3 is not triggered, how about adding a space in s3/README.md for test?

PR #1023, opened 2021-11-11
services/azblob: Migrate to Factory init.

Joey2132:
Migrate to Factory init for minio.


services/azblob/internal/cmd/meta.go
```diff
@@ -0,0 +1,135 @@
...
+			Create:       true,
+			CreateDir:    true,
+			CreateAppend: true,
+			Delete:       true,
+			List:         true,
+			Metadata:     true,
+			Read:         true,
+			Stat:         true,
+			Write:        true,
+			WriteAppend:  true,
```
Xuanwo, line 49: We need to mark `CommitAppend` as `true` here because `azblob` does support `CommitAppend`. Or, we can remove the `Appender` test for azblob.

JinnyYi, line 49: `CommitAppend` should also set to `true`.

Joey2132: > s `true` here

So how about do issue 1022 first? Or I just mark `CommitAppend` as `true` first.

Xuanwo:
Bravo!

PR #1025, opened 2021-11-12
services/s3: Update path before s.getAbsPath()

Joey2132:
Update path before s.getAbsPath() for service s3.


services/s3/storage.go
```diff
@@ -248,6 +248,19 @@ func (s *Storage) delete(ctx context.Context, path string, opt pairStorageDelete
 }
 
 func (s *Storage) list(ctx context.Context, path string, opt pairStorageList) (oi *types.ObjectIterator, err error) {
+	if !opt.HasListMode {
+		// Support `ListModePrefix` as the default `ListMode`.
+		// ref: [GSP-46](https://github.com/beyondstorage/go-storage/blob/master/docs/rfcs/654-unify-list-behavior.md)
+		opt.ListMode = types.ListModePrefix
+	}
+
+	if opt.ListMode.IsPart() {
```
Xuanwo, line 257: I think there is a typo.

Joey2132: > I think there is a typo.

Yes, so careless.

PR #1026, opened 2021-11-15
services/azfile: Migrate to Factory init

Joey2132:
services/azfile: Migrate to Factory init


services/azfile/utils.go
```diff
@@ -15,51 +15,85 @@ import (
...
+
+	defaultPairs types.DefaultServicePairs
+	features     types.ServiceFeatures
+
+	types.UnimplementedServicer
+}
+
+// String implements Servicer.String
+func (s *Service) String() string {
+	return fmt.Sprintf("Servicer azblob")
```
Xuanwo, line 31: Should be `azfile`

services/azfile/internal/cmd/meta.go
```diff
@@ -0,0 +1,67 @@
...
+		infoObjectMetaServerEncrypted,
+	},
+	Pairs: []def.Pair{},
+	Factory: []def.Pair{
+		def.PairCredential,
+		def.PairEndpoint,
+		def.PairName,
+		def.PairWorkDir,
+	},
+	Service: def.Service{
```
Xuanwo, line 20: If we don't support the service, please don't assign this service.

services/azfile/utils.go
```diff
@@ -15,51 +15,85 @@ import (
...
 	store = &Storage{
-		workDir: "/",
+		f:        *f,
+		features: f.storageFeatures(),
+		workDir:  "/",
 	}
 
-	if opt.HasWorkDir {
-		store.workDir = opt.WorkDir
+	if f.WorkDir == "" {
```
Xuanwo, line 92: If `f.WorkDir` is `""`, assign `store.WorkDir` to `""` is meaningless.

services/azfile/internal/cmd/meta.go
Joey2132: https://github.com/Joey2132/go-storage/blob/master/definitions/metadata.go#L38
If we don't assign it, the make build will fail with `panic: runtime error: invalid memory address or nil pointer dereference `.

Xuanwo: Please report a bug to go-storage before workaround it.

Joey2132: Got it.

JinnyYi:
[ListFilesAndDirectoriesSegment](https://docs.microsoft.com/en-us/rest/api/storageservices/list-directories-and-files) will list the contents only for a single level of the directory hierarchy.

[In our implementation](https://github.com/beyondstorage/go-storage/blob/master/services/azfile/storage.go#L110-L115) as follows, the current directory to list is `WorkDir`, it only lists the subdirs and files under `WorkDir`. If list with `DirMode`, I think we need `NewDirectoryURL(input.prefix)`, instead of setting `input.prefix` as `ListFilesAndDirectoriesOptions.Prefix` (Maybe we should support `List` with `ModePrefix` and `ModeDir`?).

```
options := azfile.ListFilesAndDirectoriesOptions{
	Prefix:     input.prefix,
	MaxResults: input.maxResults,
}

output, err := s.client.ListFilesAndDirectoriesSegment(ctx, input.marker, options)
```

In addition, we also need to consider path splicing, since the returned path does not contain its parent directory.

PR #1035, opened 2021-11-18
services/kodo: Migrate to Factory init

Joey2132:
services/kodo: Migrate to Factory init.


services/kodo/service.go
```diff
@@ -53,18 +56,20 @@ func (s *Service) list(ctx context.Context, opt pairServiceList) (it *typ.Storag
 }
 
 func (s *Service) nextStoragePage(ctx context.Context, page *typ.StoragerPage) error {
-	buckets, err := s.service.Buckets(false)
+	buckets, err := s.service.Buckets(true)
```
JinnyYi, line 59: Why we need to modify the `shared` to `true` here?

Joey2132: In fact, it's mistake.

PR #1038, opened 2021-11-19
services/dropbox: Migrate to Factory init

Joey2132:
services/dropbox: Migrate to Factory init.


services/dropbox/utils.go
```diff
@@ -11,11 +11,49 @@ import (
 	"go.beyondstorage.io/credential"
 	ps "go.beyondstorage.io/v5/pairs"
 	"go.beyondstorage.io/v5/services"
+	"go.beyondstorage.io/v5/types"
 	typ "go.beyondstorage.io/v5/types"
```
JinnyYi, line 15: There are duplicate references here.

Joey2132: Got it.

PR #1042, opened 2021-11-22
services/gdrive: Migrate to Factory init

Joey2132:
Migrate to Factory init.


services/gdrive/utils.go
```diff
@@ -187,7 +232,11 @@ func (s *Storage) getAbsPath(path string) string {
 		return strings.TrimPrefix(s.workDir, "/")
 	} else {
 		prefix := strings.TrimPrefix(s.workDir, "/")
-		return prefix + "/" + path
+		if !strings.HasPrefix(path, prefix) {
+			return prefix + "/" + path
+		} else {
+			return strings.TrimPrefix(path, "/")
+		}
 	}
```
JinnyYi, line 240: It looks like there is an overlap in the branches.

JinnyYi (approved):

PR #1051, opened 2021-11-23
services/minio: Migrate to Factory init.

Joey2132:
Migrate to Factory init.


services/minio/internal/cmd/meta.go
```diff
@@ -0,0 +1,87 @@
...
+	Name:        "storage_class",
+	Type:        def.Type{Name: "string"},
+	Defaultable: true,
+}
+
+// var pairExpire = def.Pair{
+// 	Name:        "expire",
+// 	Type:        def.Type{Name: "string"},
+// 	Description: "specify when the url returned by reach will expire. Deprecated: Use `QuerySignHTTPRead` instead of `Reach` and `expire` is the declared argument.",
+// }
```
JinnyYi, line 81: Maybe we should remove the no longer useful code.

Joey2132: Got it.

JinnyYi (approved):

PR #1053, opened 2021-11-23
services/ocios: Migrate to Factory init.

Joey2132:
Migrate to Factory init.


services/ocios/utils.go
```diff
@@ -20,14 +53,19 @@ func (s *Storage) String() string {
...
-	return newStorager(pairs...)
+	f := Factory{}
+	err := f.WithPairs(pairs...)
+	if err != nil {
+		return nil, err
+	}
+	return f.newStorage()
 }
 
 // newStorager will create a storage client.
```
JinnyYi, line 64: `newStorager` should be `newStorage`.

Joey2132: Got it, I will check all of them.

JinnyYi (approved):

PR #1058, opened 2021-11-25
services/oss: Migrate to Factory init.

Joey2132:
Migrate to Factory init.


services/oss/storage.go
```diff
@@ -320,6 +327,13 @@ func (s *Storage) nextObjectPageByDir(ctx context.Context, page *types.ObjectPag
 		return err
 	}
 
+	// List enpty dir will return the dir itself and we want the return value to be empty.
+	// Use current way to avoid this condition temporally.
+	// Maybe we should use ListObjectsV2 to replace ListObjects.
+	if len(output.Objects) == 1 && output.Objects[0].Key == input.prefix && output.Objects[0].Size == 0 {
+		return types.IterateDone
+	}
+
```
JinnyYi, line 336: Maybe we don't need the weird processing to try to pass the test? `TestListEmptyDir` has been updated and will no longer create an empty directory for services that don't natively support dir.

JinnyYi (approved):

PR #1060, opened 2021-11-29
services/tar: Migrate to Factory init.

Joey2132:
Migrate to Factory init.


services/tar/internal/cmd/meta.go
```diff
@@ -0,0 +1,54 @@
...
+	Storage: def.Storage{
+		Features: types.StorageFeatures{
+			WriteEmptyObject: true,
+
+			Delete:   true,
+			List:     true,
+			Metadata: true,
+			Read:     true,
+			Stat:     true,
+			Write:    true,
```
JinnyYi, line 28: In fact, only `List`, `Read` and `Stat` are currently implemented.

Joey2132: Got it, I will check it.

services/tar/internal/cmd/meta.go
```diff
@@ -20,12 +20,9 @@ var Metadata = def.Metadata{
 		Features: types.StorageFeatures{
 			WriteEmptyObject: true,
```
JinnyYi, line 21: `WriteEmptyObject` should also be `false`.

services/tar/utils.go
```diff
@@ -12,10 +12,44 @@ import (
...
+
+	defaultPairs types.DefaultServicePairs
+	features     types.ServiceFeatures
+
+	types.UnimplementedServicer
+}
+
+// String implements Servicer.String
+func (s *Service) String() string {
+	return fmt.Sprintf("Servicer cephfs")
```
JinnyYi, line 28: Service name should be `tar`.

JinnyYi (approved):

PR #1062, opened 2021-11-29
services/us3: Migrate to Factory init.

Joey2132:
Migrate to Factory init.


services/us3/internal/cmd/meta.go
```diff
@@ -0,0 +1,55 @@
...
+		Features: types.StorageFeatures{
+			WriteEmptyObject: true,
+
+			Create:   true,
+			Delete:   true,
+			Metadata: true,
+			List:     true,
+			Read:     true,
+			Stat:     true,
+			Write:    true,
```
JinnyYi, line 28: Maybe we should not change the default values of the features before implementing them?

JinnyYi (approved):

PR #1064, opened 2021-11-29
services/zip: Migrate to Factory init

Joey2132:
Migrate to Factory init.


services/zip/internal/cmd/meta.go
```diff
@@ -0,0 +1,50 @@
...
+	Storage: def.Storage{
+		Features: types.StorageFeatures{
+			WriteEmptyObject: true,
+
+			Create:   true,
+			Delete:   true,
+			List:     true,
+			Metadata: true,
+			Read:     true,
+			Write:    true,
```
JinnyYi, line 25: Maybe we should keep the default values of the features before implementing them?

JinnyYi (approved):

PR #1092, opened 2021-12-24
fix(services/minio): Set optional location when creating service

houz42:


services/minio/internal/cmd/meta.go
```diff
@@ -26,6 +27,16 @@ var Metadata = def.Metadata{
 			Get:    true,
 			List:   true,
 		},
+
+		Create: []def.Pair{
+			def.PairLocation,
+		},
+		Delete: []def.Pair{
+			def.PairLocation,
```
JinnyYi, line 35: Maybe `Location` is not a necessary parameter for `RemoveBucket`?

houz42: No, not always necessary, but it's optional. Should optional parameter be here, or somewher else?

JinnyYi: The optional parameters are indeed specified here. There are no optional parameters for `RemoveBucket` at this time from the [doc](https://docs.min.io/docs/golang-client-api-reference.html#RemoveBucket)?

services/minio/service.go
```diff
@@ -15,7 +15,7 @@ func (s *Service) create(ctx context.Context, name string, opt pairServiceCreate
 	if err != nil {
 		return nil, err
 	}
-	err = s.service.MakeBucket(ctx, name, minio.MakeBucketOptions{})
+	err = s.service.MakeBucket(ctx, name, minio.MakeBucketOptions{Region: opt.Location})
```
JinnyYi, line 18: Maybe we also need to set `f.Location = opt.Location` above so that we can use `f.Location` in `utils.go`?

PR #1149, opened 2022-03-02
Fix IterateDone error type on Readme

kondanta:
Hi 👋🏼 

I've realized that `types.IteratorDone` is not available in the current API, as mentioned in README.


README.md
```diff
@@ -73,7 +73,7 @@ func main() {
     for {
     	// Use iterator.Next to retrieve next object until we meet IteratorDone.
     	o, err := it.Next()
-    	if errors.Is(err, types.IteraoorDone) {
+    	if errors.Is(err, types.IterateDone) {
```
JinnyYi, line 76: Can you fix the errors in the comments and Chinese document together?

xxchan:
Thanks!

xxchan (approved):

README.md
kondanta: I've updated the comment line in question along with the Chinese read-me. In Chinese one, I've just updated the type, though.

JinnyYi (approved):

